PR #45, opened 2022-09-30
Build base Fatman images by image-builder

iszulcdeepsense:


image_builder/Dockerfile
```diff
@@ -13,12 +13,12 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /
   "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/debian \
   $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null &&\
     apt-get update -y && apt-get install -y docker-ce-cli
+# apt cache is cleaned automatically, see /etc/apt/apt.conf.d/docker-clean
```
Koshmaar, line 16: Useful comment, without it people could try to add `apt clean` at end.

image_builder/image_builder/docker/builder.py
```diff
@@ -40,27 +39,44 @@ def build(
...
+                logger.info(f'base image not found in a registry, rebuilding {base_image}, '
+                            f'deployment ID: {deployment_id}, keeping logs in {base_logs_filename}')
+                build_container_image(
+                    base_image,
+                    base_image_path,
+                    base_image_path.parent,
+                    metric_labels,
+                    base_logs_filename,
+                )
+                logger.info(f'base Fatman image has been built and pushed: {base_image}')
```
Koshmaar, line 70: Nitpick: there's quite a lof happening in this function `build`, here and below. Perhaps 1-2 parts could be extracted to separate functions (only if it makes sense).

image_builder/image_builder/docker/builder.py
```diff
@@ -116,30 +132,34 @@ def _wait_for_docker_engine_ready():
...
+                job_name = name_parts[0]
+                job_version = name_parts[1] if len(name_parts) > 1 else 'latest'
+                job_templates[job_name] = (base_image_path, template_path, job_version)
 
     return job_templates
+
+
+def _image_exists_in_registry(image_name: str):
+    try:
+        shell(f'docker manifest inspect --insecure {image_name}')
```
Koshmaar, line 160: Ahh, so that's clever trick to check whether image exists on docker hub - that was the intention here, right?  Didn't know it. If it was only about localhost docker registry, `docker image inspect` would do the trick. One downside of `docker manifest` that it takes about 10 sec to check for image that exists, and 5 for one that doesn't exist. Btw, it's experimental feature in docker, so worth adding comment about it.

Koshmaar (approved):

image_builder/image_builder/docker/builder.py
iszulcdeepsense: Yes, it's supposed to work with localhost registry as well as docker hub / github registry or any on-premises Gitlab registry.
It's true that sometimes it can be slow, though it's the matter of 20ms to check if image exists in local registry on my computer. For ghcr.io registry it lasts about 3 seconds.

If you know a better way to do it, I'll appreciate that.

PR #46, opened 2022-09-30
Support multiple versions of the same plugin

iszulcdeepsense:
Support multiple versions of the same plugin

Set proper permissions of plugin files

Check if user is really allowed to manage plugins

### DIFF
--- a/lifecycle/lifecycle/auth/check.py
+++ b/lifecycle/lifecycle/auth/check.py
@@ -104,6 +104,6 @@ def check_staff_user(
     if not user.is_active:
         raise UnauthorizedError(f'User {username} is not active')
     if not user.is_staff:
-        raise UnauthorizedError(f'User {username} is a staff member')
+        raise UnauthorizedError(f'User {username} is not a staff member')
 
     return auth_subject
--- a/lifecycle/lifecycle/auth/check.py
+++ b/lifecycle/lifecycle/auth/check.py
@@ -1,6 +1,9 @@
 from functools import wraps
 from typing import List
 
+from django.contrib.auth.models import User
+from django.http import Http404
+from django.shortcuts import get_object_or_404
 from fastapi import Request
 from fastapi.responses import JSONResponse
 
@@ -80,3 +83,27 @@ def check_auth_decorator(
 
         return decorated
     return auth_decorator
+
+
+def check_staff_user(
+    request: Request,
+) -> models.AuthSubject:
+    """
+    Authenticate and authorize the request,
+    checking if the request is being made by a staff user.
+    :raise UnauthorizedError: If auth subject has no sufficient permissions
+    """
+    token_payload, auth_subject = authenticate_token(request)
+    authorize_subject_type(token_payload, [AuthSubjectType.USER])
+    username = token_payload.subject
+
+    try:
+        user: User = get_object_or_404(User, username=username)
+    except Http404:
+        raise UnauthorizedError(f'User {username} was not found')
+    if not user.is_active:
+        raise UnauthorizedError(f'User {username} is not active')
+    if not user.is_staff:
+        raise UnauthorizedError(f'User {username} is a staff member')
+
+    return auth_subject
--- a/racetrack_commons/racetrack_commons/plugin/engine.py
+++ b/racetrack_commons/racetrack_commons/plugin/engine.py
@@ -72,11 +72,19 @@ class PluginEngine:
             logger.warning(f'Plugin {plugin_name} does not have hook {function_name}')
             return None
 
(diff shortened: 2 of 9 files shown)


lifecycle/lifecycle/auth/check.py
```diff
@@ -80,3 +83,27 @@ def decorated(*args, **kwargs):
...
+    username = token_payload.subject
+
+    try:
+        user: User = get_object_or_404(User, username=username)
+    except Http404:
+        raise UnauthorizedError(f'User {username} was not found')
+    if not user.is_active:
+        raise UnauthorizedError(f'User {username} is not active')
+    if not user.is_staff:
+        raise UnauthorizedError(f'User {username} is a staff member')
```
Koshmaar, line 107: Typo: is not a staff member.

racetrack_commons/racetrack_commons/plugin/engine.py
```diff
@@ -72,11 +72,19 @@ def invoke_one_plugin_hook(self, plugin_name: str, function: Callable, *args, **
...
+                if plugin_data.plugin_manifest.name == name and plugin_data.plugin_manifest.version == version:
+                    return plugin_data
+            raise EntityNotFound(f'plugin named "{name}" {version} was not found')
+
+        else:
+            plugin_versions = [plugin_data for plugin_data in self.plugins_data
+                               if plugin_data.plugin_manifest.name == name]
+            if not plugin_versions:
+                raise EntityNotFound(f'plugin named "{name}" was not found')
+            return plugin_versions[-1]
```
Koshmaar, line 87: How do you know that the last one will be.. newest? Don't see any sort. What's the logic here anyway, return any version, or newest?

racetrack_commons/racetrack_commons/plugin/loader.py
```diff
@@ -31,28 +32,46 @@ def load_plugins_from_dir(plugins_dir: str) -> List[PluginData]:
...
 
 
 def load_plugin_from_zip(plugin_zip_path: Path) -> PluginData:
     assert plugin_zip_path.is_file(), f'no such file {plugin_zip_path}'
-    extracted_plugin_path = plugin_zip_path.parent / EXTRACTED_PLUGINS_DIR / plugin_zip_path.stem
     
+    extracted_plugins_dir = plugin_zip_path.parent / EXTRACTED_PLUGINS_DIR
+    if not extracted_plugins_dir.is_dir():
+        extracted_plugins_dir.mkdir(parents=True, exist_ok=True)
+        extracted_plugins_dir.chmod(mode=0o777)  # mkdir(mode=0o777) doesn't really work
```
Koshmaar, line 46: What it means doesnt' work? Leftover comment? Or if not, if 777 doesnt' work, should it be changed?

Koshmaar (approved):

racetrack_commons/racetrack_commons/plugin/engine.py
iszulcdeepsense: It is supposed to be sorted by `load_plugins_from_dir` function from here: https://github.com/TheRacetrack/racetrack/blob/006fcdc136855d4f0377af46cfabf25fe93289b9/racetrack_commons/racetrack_commons/plugin/loader.py#L36
The logic is to take the newest plugin. It's sorted by "priority, then by version ascending" so that the most important plugins are invoked at the end in case they would interfere with each other. I can't find the real case now, but who knows what plugins will be created in future? 

It's not solving any problem right now, but I think it's good to have a deterministic order of execution at least. So it's rather to secure for the future.

racetrack_commons/racetrack_commons/plugin/loader.py
iszulcdeepsense: I wrote a comment in case if anyone wonders why I didn't do it in one line: `extracted_plugins_dir.mkdir(parents=True, exist_ok=True, mode=0o777)`

racetrack_commons/racetrack_commons/plugin/engine.py
Koshmaar: Ok, didn't get it from reading this function that the list will be sorted.

racetrack_commons/racetrack_commons/plugin/loader.py
Koshmaar: ah, so I got confused by comment, and thought it means `chmod(777)` doesn't work. I think the comment should be on mkdir line, or removed at all, as it's comment about python's file API.

iszulcdeepsense: Alright, I'll get rid of it.

Issue #37, opened 2022-09-29
Plugins Hub


iszulcdeepsense:
Let's have a one place listing all the public plugins available for Racetrack.
It can be a simple Markdown page with links to the "Releases" pages, which allows to download ZIP plugins in a semi-automated way.

Additionally, there can be a `racetrack install <plugin_name> <racetrack_url>` command which downloads the ZIP plugin and uploads it to a given Racetrack instance.

Moreover, there can be a dedicated page conducive to pick the plugins by an admin and install (or upgrade) them in a convenient manner.

Maybe it would be useful do to racetrack plugin list <racetrack_url>

PR #56, opened 2022-10-04
Preinstall python plugin

iszulcdeepsense:


racetrack_commons/racetrack_commons/plugin/engine.py
```diff
@@ -129,6 +130,18 @@ def on_moved(self, event):
...
+        # inotify may not work on network filesystems
+        def _check_periodically():
+            while True:
+                time.sleep(5 * 60)
+                try:
+                    compare_changes()
+                except BaseException as e:
+                    log_exception(ContextError('periodic check failure', e))
+
+        threading.Thread(target=_check_periodically, daemon=True).start()
```
Koshmaar: Is this thread solely for checking that the (network) disk can be read? Wouldn't such problem be visible from the start?

Btw, I already reviewed this, but now forgot how it works - there are many handlers (on_moved etc) that start thread with compare_changes in deamon mode. Won't this cause threads to be created and never finished?

racetrack_client/racetrack_client/main.py
```diff
@@ -138,6 +139,23 @@ def _print_help(_: argparse.Namespace):
     parser_logout.add_argument('racetrack_url', default='', nargs='?', help='URL to Racetrack server or alias name')
     parser_logout.set_defaults(func=_logout)
 
+    # racetrack plugin
+    parser_plugin = subparsers.add_parser('plugin', help='Manage Racetrack plugins')
+    subparsers_plugin = parser_plugin.add_subparsers()
```
Koshmaar, line 144: Maybe it would be useful do to `racetrack plugin list <racetrack_url>`. Just an idea for future.

racetrack_commons/racetrack_commons/plugin/engine.py
```diff
@@ -129,14 +130,29 @@ def on_moved(self, event):
...
+    def upload_plugin(self, filename: str, file_bytes: bytes) -> PluginManifest:
         target_zip = Path(self.plugins_dir) / filename
         assert target_zip.suffix == '.zip', '.zip plugins are only supported'
 
         # save to tmp.zip to avoid overwriting current plugins
         tmp_zip = Path(self.plugins_dir) / 'tmp.zip'
+        if not tmp_zip.is_file():
+            tmp_zip.touch()
+            tmp_zip.chmod(mode=0o666)
         tmp_zip.write_bytes(file_bytes)
```
Koshmaar, line 154: I guess that when tmp.zip already exists, then write_bytes will overwrite it with new content?

docs/user.md
```diff
@@ -651,7 +653,8 @@ source venv/bin/activate
 docker system prune -a
 make kind-up
 # Wait 10 minutes
-make test
+make kind-test
+racetrack plugin install https://github.com/TheRacetrack/plugin-python-job-type/releases/download/2.4.0/python3-job-type-2.4.0.zip http://localhost:7002
```
Koshmaar, line 657: Won't the python3 job type be needed for kind-test deployment tests? 

Another thing - how about adding the `racetrack plugin install ... ` to make kind-test (and perhaps other Makefile cmds) so that whenever developer runs it, he has a fully functional setup (like it was in my days ;) ). But not insisting, as it's just minor automation, and I'm not 100% sure if that makes sense with the flexible plugin architecture.

Koshmaar (approved):

racetrack_commons/racetrack_commons/plugin/engine.py
iszulcdeepsense: > Is this thread solely for checking that the (network) disk can be read? Wouldn't such problem be visible from the start?

NFS disk can be read for sure, but for some reasons it doesn't work with inotify. I'm not sure how to check that, but I suppose it might be quite complex.

> Won't this cause threads to be created and never finished?

That's exactly their purpose - to never finish and to monitor plugins until the end of the container's life.

racetrack_client/racetrack_client/main.py
iszulcdeepsense: Good idea. It could be part of this issue https://github.com/TheRacetrack/racetrack/issues/37

docs/user.md
iszulcdeepsense: The plugin is needed indeed. Good news, I've already added it to test_deploy.py so it's always being installed as a part of end-to-end test (either on docker-compose or kind). However the `racetrack plugin install` is rather needed before `racetrack deploy` command, I left it in order not to depend on `kind-test` result

Incidentally, I'm hoping to have fully functional setup easily too. I'm planning to improve that flow when doing https://github.com/TheRacetrack/racetrack/issues/37. Right now, it was more like a first step to make it working (and pass tests on CI). Stay tuned :)

racetrack_commons/racetrack_commons/plugin/engine.py
iszulcdeepsense: Do you think it's something be should bother?

Koshmaar: It's not something likely to happen, but deleting the file upfront, plus generating few chars of random name ie. tmp_123456.zip makes this situation impossible. One less problem to worry about or debug afterwards :) and it's 3 lines of code.

racetrack_commons/racetrack_commons/plugin/engine.py
Koshmaar: > NFS disk can be read for sure, but for some reasons it doesn't work with inotify. I'm not sure how to check that, but I suppose it might be quite complex. Hence the safety check 

Ah, now I see. So it's additional "manual" check that is run in addition to the ChangesHandler (which is using ionotify). Frankly, if ionotify can't be trusted, and you have to add additional layer just to be sure... I'd just remove ionotify. Cause it's quite some code which is not trivial, and it duplicates with the manual check.

> That's exactly their purpose - to never finish and to monitor plugins until the end of the container's life.

Still don't understand why it won't ultimately cause hundreds of thousands of threads to be spinning in the background, watching the changes :] as every time the LAST_CHANGE_FILE is touched in any way, neverending thread is started.

docs/user.md
Koshmaar: Ok, so `racetrack plugin install` is done also inside `kind-test`, but you don't know if that will always happen in the future, that's why you're adding it here just to be sure user will have it working. Fair enough.

racetrack_commons/racetrack_commons/plugin/engine.py
iszulcdeepsense: > Still don't understand why it won't ultimately cause hundreds of thousands of threads to be spinning in the background, watching the changes :] as every time the LAST_CHANGE_FILE is touched in any way, neverending thread is started.

Now I see the misunderstanding here. The thread `threading.Thread(target=_check_periodically, daemon=True).start()` is spawned only once, inside `_watch_plugins_changes` function, which is being run at startup, in a `PluginEngine` constructor.

While the other threads like this
```python
        def on_created(self, event):
            if Path(event.src_path).name == LAST_CHANGE_FILE:
                logger.debug(f'file creation notified on {event.src_path}')
                threading.Thread(target=compare_changes, daemon=True).start()
```
are only checking for changes and they finish right after it, so it's not "neverending".

I wanted to keep the observer thread alive in case of error so I moved that to separate threads, but now I see I could just keep it inside try-catch block.

>  if ionotify can't be trusted, and you have to add additional layer just to be sure... I'd just remove ionotify

But if inotify works properly, it's better to have the plugins reloaded immediately right after the change, without the need to wait for the next periodic check cycle.

racetrack_commons/racetrack_commons/plugin/engine.py
iszulcdeepsense: Right, random name is a good way to go.

racetrack_commons/racetrack_commons/plugin/engine.py
Koshmaar: > But if inotify works properly, it's better to have the plugins reloaded immediately right after the change, without the need to wait for the next periodic check cycle.

You could set the period to 10 or 15 sec, which from the perspective of user uploading file on server is almost instant, and it's very light on server. Your decision :)

iszulcdeepsense: It's tough choice. If it's working with Network Filesystem volume, then it's maybe not so healthy to ping it every 10 seconds due to additional requests over the network.

PR #70, opened 2022-10-19
Allow users to change their password

iszulcdeepsense:


dashboard/dashboard/views.py
```diff
@@ -51,17 +52,34 @@ def clean_username(self):
         return username
 
 
+class RacetrackPasswordChangeForm(PasswordChangeForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for visible in self.visible_fields():
+            visible.field.widget.attrs['class'] = 'form-control'
```
Koshmaar, line 59: Hmm, why is this needed?

dashboard/dashboard/templates/registration/change_password.html
```diff
@@ -0,0 +1,37 @@
...
+
+                {% for error in field.errors %}
+                    <div class="invalid-feedback d-block">
+                    {{ error|escape }}
+                    </div>
+                {% endfor %}
+            </div>
+        {% endfor %}
+
+        <button type="submit" class="btn btn-primary">Change Password</button>
```
Koshmaar, line 30: Recently at workshop I learned about zxcvbn - it's a password strength validation tool, to encourage users to use more secure passwords. Because with simple rules like use big letter, use number, use special char, we end up with passwords like "Password1!". Hackers know this, and setup their brute force tools accordingly, and passwords can be cracked in a matter of minutes or hours (instead of days or months, like it should be). So https://github.com/dropbox/zxcvbn was invented to improve on that, and https://github.com/Pawamoy/django-zxcvbn-password is django plugin (looking reasonably supported) with nice bar showing how long it will take to crack used password. So it's a UI eye-candy, that encourages user to choose more secure password. If you have the time, might be worth installing it (perhaps separate PR).

Django zxcvbn

Koshmaar (approved):

dashboard/dashboard/templates/registration/change_password.html
```diff
@@ -0,0 +1,37 @@
+{% extends "racetrack/base.html" %}
+
+{% block content %}
+
+<div class="container">
+<div class="card">
+<div class="card-body">
+
+    <h5 class="card-title">Change Password</h5>
```
Koshmaar, line 9: Ah, you could also display here password rules, so users upfront know what is the minimum length, special chars etc. without having to go through trial-error.

dashboard/dashboard/views.py
iszulcdeepsense: This adds `class` attribute to HTML form elements, giving them a bootstrap style.

dashboard/dashboard/templates/registration/change_password.html
iszulcdeepsense: Fortunately, it is displayed out-of-the-box thanks to default Django's password form :smile: 

Screenshot from 2022-10-25 15-58-44

Koshmaar: Great :) forgot about it.

dashboard/dashboard/views.py
Koshmaar: Ok, can't it be done in static html? ie. I see there `class="form-text"`

iszulcdeepsense: The problem is that form components are created dynamically in a template (it's part of `change_password.html`)
```
{% for field in form %}
   ...
                {{ field }}
   ...
{% endfor %}
```
so I can't set the class in a static manner. There are more advantages though, the proper components are created automatically by Django based on a default form for changing the password.

dashboard/dashboard/templates/registration/change_password.html
iszulcdeepsense: Cool, it might be noteworthy. Though in this case, having so many restrictive (annoying :smile: ) rules provided by Django: "password must contain at least 8 characters, can't be a commonly used password, can't be entirely numeric" there's a higher chance of ending up with the strong password either way.

PR #79, opened 2022-11-07
Add debugging tools to racetrack images

iszulcdeepsense:
`dnsutils` and `curl` binaries are pretty useful when debugging DNS / network issues inside containers. Since the containers run on non-root user, it comes in handy to ship the images with these packages.

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -2,8 +2,9 @@ FROM python:3.9-slim-bullseye
 
 WORKDIR /src/dashboard
 
-RUN apt-get update -y && apt-get install -y postgresql-client \
-  curl dnsutils vim
+RUN apt-get update -y && apt-get install -y \
+  postgresql-client curl dnsutils vim &&\
+  rm -rf /var/lib/apt/lists/*
 # apt cache is cleaned automatically, see /etc/apt/apt.conf.d/docker-clean
 
 COPY racetrack_client/setup.py racetrack_client/requirements.txt racetrack_client/README.md /src/racetrack_client/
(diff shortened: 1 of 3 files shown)


dashboard/Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM python:3.9-slim-bullseye
 
 WORKDIR /src/dashboard
 
-RUN apt-get update -y && apt-get install -y postgresql-client
+RUN apt-get update -y && apt-get install -y postgresql-client \
```
Koshmaar, line 5: It's recommended to clear the apt cache at the end of update: `&& rm -rf /var/lib/apt/lists/*` , here and in other images also.

See: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

Koshmaar (approved):

dashboard/Dockerfile
iszulcdeepsense: I think you're right. Once I skipped that on purpose due to 
> Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

which is included in /etc/apt/apt.conf.d/docker-clean in python:3.9-slim-bullseye image.

but now I think `apt-get clean` is something different `rm -rf /var/lib/apt/lists/*` so I probably need both.

Koshmaar: Yeah, judging by this code: https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105 , the apt cache clean is not cleaning /apt/lists/.

PR #81, opened 2022-11-07
Evaluate database name based on the cluster name

iszulcdeepsense:


racetrack_commons/racetrack_commons/database/database.py
```diff
@@ -0,0 +1,41 @@
+import os
+from pathlib import Path
+from typing import Dict
+
+
+def populate_database_settings(base_dir: Path) -> Dict[str, Dict]:
```
Koshmaar, line 6: Since it's quite long function, how about adding unit tests for the logic of constructing database_name? That could be crudely done manipulating os.env['POSTGRES_DB'] etc, or extracting that part as function.

Koshmaar (approved):

racetrack_commons/racetrack_commons/database/database.py
iszulcdeepsense: Right. I've added unit test

PR #89, opened 2022-11-15
Add "job type version" column to portfolio table

iszulcdeepsense:


dashboard/dashboard/templates/racetrack/portfolio.html
```diff
@@ -24,6 +24,7 @@ <h5 class="card-title">Fatmen Portfolio</h5>
                         <th>Family name</th>
                         <th>Version</th>
                         <th>Status</th>
+                        <th>Job type version</th>
```
Koshmaar, line 27: Fatman has its version, and there's version of its... job type. That's confusing for me (maybe I'm too far from RT now :) ), and for less technical users will probably also be. "What's a job within fatman?" they ask. Maybe you could add a on-hover description on how to simply understand that term.

Koshmaar (approved):

dashboard/dashboard/views.py
```diff
@@ -191,6 +192,8 @@ def view_fatman_portfolio(request):
         for fatman_dict in fatmen_dicts:
             fatman_dict['update_time_days_ago'] = days_ago(fatman_dict['update_time'])
             fatman_dict['last_call_time_days_ago'] = days_ago(fatman_dict['last_call_time'])
+            manifest: Dict = fatman_dict.get('manifest')
+            fatman_dict['job_type_version'] = manifest.get('lang') if manifest else None
```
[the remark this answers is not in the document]
iszulcdeepsense: Good point, the import is not needed now. It turned out what's inside is not the `Manifest` type, but the dict.

dashboard/dashboard/templates/racetrack/portfolio.html
iszulcdeepsense: Agree, it might be confusing. WIll add the tooltip and I'll rename `Version` to `Fatman version`

PR #91, opened 2022-11-16
Add documentation about fine-grained permission model

iszulcdeepsense:


docs/permissions.md
```diff
@@ -0,0 +1,88 @@
+# Permission model
+Racetrack v2 introduces a restrictive Fatman permissions model.
+
+**Users**, **ESCs** and **Fatman Families** can be granted fine-grained permissions.
+For instance, users can have access to a subset of fatmen and they
+can see and manage only those that are allowed for them.
+
+By default, nobody is allowed to do anything (if the permissions list is empty).
+However, a brand new user comes with a default set of permissions allowing him to:
```
Koshmaar, line 9: What do you mean by "brand new user" - new one created freshly by admin? I mean, in the context of "nobody is allowed to do anything", I understand that the users list is non-empty, but nobody can do anything - so system looks unsusable, even though it has users. But now comes the new user who (surprise) by default has some permissions. That could be cleared.

docs/permissions.md
```diff
@@ -0,0 +1,88 @@
...
+- Fatman family
+- Fatman
+- Endpoint
+
+All of these fields are optional. Each of them narrows down the filtering of the resources, which the permission covers.
+For instance, if **Fatman** field is empty and the **Family** is set, the permission works for all fatmen within the family.
+If none of the fields is filled, the permission works on all of the resources (all families, all fatmen, all endpoints).
+
+Keep in mind that the permission may give an access not only to the existing resources, 
+but also for those that are yet to come up in future
```
Koshmaar, line 67: Nitpick: lack of dot at the end :)

Koshmaar (approved):

docs/permissions.md
iszulcdeepsense: Indeed, it's confusing. Let me explain.

> nobody is allowed to do anything (if the permissions list is empty)

That's true regarding **Fatman** permissions. Users can still log in, admin can add data models, so system is usable by admin. Django permissions are orthogonal.
With this sentence I wanted to note that the access is denied by default (unless there is a rule allowing to do something).

When a new user is created, a set of default permissions is created accordingly. 
So the permissions list is not empty any longer and that invalidates statement "nobody is allowed to do anything".

The permissions of a user can be revoked by admin later on. That's why it refers to "new users".

I'll rephrase that in docs.

PR #97, opened 2022-11-25
List available job type versions

iszulcdeepsense:
Screenshot from 2022-11-25 15-40-59

### DIFF
--- a/image_builder/image_builder/job_type.py
+++ b/image_builder/image_builder/job_type.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+from collections import defaultdict
+from dataclasses import dataclass
+from pathlib import Path
+
+import backoff
+
+from racetrack_client.log.context_error import wrap_context
+from racetrack_client.utils.semver import SemanticVersion
+from racetrack_commons.plugin.core import PluginCore
+from racetrack_commons.plugin.engine import PluginEngine
+
+
+@dataclass
+class JobTypeVersion:
+    lang_name: str
+    version: str  # semantic version of the job type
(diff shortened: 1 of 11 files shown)


racetrack_client/racetrack_client/plugin/install.py
```diff
@@ -80,6 +80,23 @@ def list_installed_plugins(lifecycle_url: str):
         print(info)
 
 
+def list_available_job_types(lifecycle_url: str):
```
Koshmaar, line 83: Nitpick: missing return type `-> None`.

image_builder/image_builder/job_type.py
```diff
@@ -0,0 +1,75 @@
...
+import backoff
+
+from racetrack_client.log.context_error import wrap_context
+from racetrack_client.utils.semver import SemanticVersion
+from racetrack_commons.plugin.core import PluginCore
+from racetrack_commons.plugin.engine import PluginEngine
+
+
+@dataclass
+class JobTypeVersion:
```
Koshmaar, line 15: I have ambigous feeling for the name JobType**Version** - looking at the class, it looks more like JobType. Don't see reason to emphasize that it stores version... it stores other fields too. It's visible also in the names of the functions below - `load_job_type`, `gather_job_types`. Food for thought :)

Koshmaar (approved):

image_builder/image_builder/job_type.py
iszulcdeepsense: You got me. I've noticed I'm using it interchangeably `job_type: JobTypeVersion`, so it definitely makes sense to call it `JobType`

PR #105, opened 2022-12-01
Add note about conflicting Fatman's requirements

iszulcdeepsense:


docs/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `python3` base image has now `git` and `build-essential` installed (again).
 - Plugin's requirements are installed with non-root user context.
 - Plugin directory is accessible when initializing the plugin in `__init__` method.
+- User-defined requirements, coming from `python3` jobs,
+  are better isolated from the core fatman dependencies to avoid conflicts.
```
JosefAssadERST, line 20: Wouldn't this be a change in the plugin though?

docs/development/plugins-job-types.md
```diff
@@ -286,7 +286,8 @@ Every wrapper has to follow some rules:
 - You MAY implement swagger documentation for your endpoints on root endpoint.
 - You MAY implement `/metrics` endpoint for exposing Prometheus metrics.
 - You MAY expose any other endpoints.
-
+- Be careful to isolate libraries / requirements installed by the user
+  from the versions of the libraries used by the core wrapper.
```
JosefAssadERST, line 290: Good phrasing.

docs/CHANGELOG.md
iszulcdeepsense: Well, there's a little muddle with the docs since splitting python3 from racetrack. Though docs are still here, I think it's more convenient to read the one Changelog. Also, it's easier to navigate docs pages (like [this](https://theracetrack.github.io/racetrack/docs/job_types/job_python3/)) if it's in one place.
Do you think we should have separate changelogs?

JosefAssadERST: In this case it's a change in the job type, so yeah.

Like for example if we produce a Haskell job type today, and someone begins to use it. Then tomorrow we change how it behaves, it's not a change in Racetrack. It's a change in the job type. It can be hard to spot because the Python job type is the foundational one, the one that started it all and was part of core. This change is still part of the job type though, it seems to me.

iszulcdeepsense: Yep, it makes sense to split out the Changelogs

PR #110, opened 2022-12-12
Regenerate auth tokens on demand

iszulcdeepsense:
When you restore the snapshot of the Racetrack's database originating from the other environment (with other `AUTH_KEY`), it can break down the tokens, making the signature invalid with the current `AUTH_KEY`. Thus, we need an endpoint for admins to recreate the signatures of all tokens.

In particular, migration from v1 to v2 goes through the `lifecycle/lifecycle/django/registry/migrations/0021_migrate_permissions.py` which depends on the current value of `AUTH_KEY`. So if it was changed after the migration, it might cause a malfunction.

### DIFF
--- a/docs/admin.md
+++ b/docs/admin.md
@@ -156,6 +156,17 @@ tool to restore the database.
 Otherwise, if your database runs inside kubernetes, 
 exec to `postgres` pod and use [pg_restore](https://www.postgresql.org/docs/current/app-pgrestore.html). 
 
+##### Rotating Auth Key
+If you restore a snapshot of the Racetrack's database originating from the other environment (with different `AUTH_KEY`),
+it can invalidate the auth tokens, making the signature invalid with the current `AUTH_KEY`.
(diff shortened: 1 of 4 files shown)


Koshmaar:
> When you restore the snapshot of the Racetrack's database originating from the other environment (with other `AUTH_KEY`), it can break down the tokens, making the signature invalid with the current `AUTH_KEY`. Thus, we need an endpoint for admins to recreate the signatures of all tokens.
> 
> In particular, migration from v1 to v2 goes through the `lifecycle/lifecycle/django/registry/migrations/0021_migrate_permissions.py` which depends on the current value of `AUTH_KEY`. So if it was changed after the migration, it might cause a malfunction.

I think it would be good to add such comment in administration.md (or any other similar file), in Backups/Migrations part, so that users know when and how to use this functionality.

Koshmaar (approved):

PR #114, opened 2022-12-21
Support multiple containers provided by job type plugins

iszulcdeepsense:
RIght now, job type plugins can't make full use of fatmen composed of multiple containers. Although there is one Dockerfile-base job type, this Dockerfile always comes from a manifest (given by user), and there is no possibility to provide it by job type itself.

This is also needed to support HUGO job type


racetrack_commons/racetrack_commons/deploy/job_type.py
```diff
@@ -49,14 +49,30 @@ def gather_job_types(
...
+
+                assert len(base_image_paths) > 0, 'Job type should have non-empty list of base'
+
+                for i, base_image_path in enumerate(base_image_paths):
+                    template_path = template_paths[i]
+                    if base_image_path is not None:
+                        assert base_image_path.is_file(), f'cannot find base image Dockerfile for {job_full_name} language wrapper: {base_image_path}'
+                    if template_path is not None:
+                        assert template_path.is_file(), f'cannot find Dockerfile template for {job_full_name} language wrapper: {template_path}'
+
```
Koshmaar, line 70: Nitpick: putting these to separate validation function would make for shorter/simpler main function code.

racetrack_commons/racetrack_commons/deploy/image.py
```diff
@@ -1,27 +1,13 @@
...
-    return get_full_image(docker_registry, registry_namespace, 'fatman-entrypoint', name, tag)
-
-
-def get_fatman_user_module_image(docker_registry: str, registry_namespace: str, name: str, tag: str) -> str:
-    """Return full name of Fatman user-module image (from Dockerfile job type)"""
-    return get_full_image(docker_registry, registry_namespace, 'fatman-user-module', name, tag)
+def get_fatman_image(docker_registry: str, registry_namespace: str, name: str, tag: str, module_index: int = 0) -> str:
+    """
+    Return full name of Fatman image.
+    If a fatman is composed of multiple containers, the module number is included.
```
Koshmaar, line 7: The `module` parameter is not easily understandable in the context of images and containers.. just complaining, don't have good proposition here. Maybe more explanation what user should pass here, what are the expectations, what kind of logic it is modelling, would help. Btw, in the DockerBuilder::build you pass there `image_index` param.

docs/development/developing-plugins.md
```diff
@@ -80,10 +80,10 @@ def fatman_runtime_env_vars(self) -> Optional[Dict[str, str]]:
 
 - `fatman_job_types` - Job types provided by this plugin
 ```python
-def fatman_job_types(self) -> Dict[str, Tuple[Path, Path]]:
+def fatman_job_types(self) -> dict[str, list[tuple[Path, Path]]]:
```
Koshmaar, line 83: Thinking out loud... this type is not very friendly, and it's used in many places. Yes, more complex are possible. Still it's a bit to read and analyze. What do you think about "typedefing" it (frequent C++ practice) and using through-out, ie. 
```
DictListImagesPath = dict[str, list[tuple[Path, Path]]]

def fatman_job_types(self) -> DictListImagesPath:
     ...
```
TBH, I'm not very proud of proposed name, maybe there's something more clever here.

Koshmaar (approved):

docs/development/developing-plugins.md
iszulcdeepsense: Yeah, I think I could use a typedef. However, this function is a public interface of the Plugin class, so I think it's good that you don't need to import any types when creating your plugin:
```python
from __future__ import annotations
from pathlib import Path

class Plugin:
    def fatman_job_types(self) -> dict[str, list[tuple[Path, Path]]]:
```
so that plugins remain independent of the Racetrack codebase.

Surely, I agree this isn't a friendly type, so I'll use a typedef at least in the internal module `job_type.py`. Is it a good trade-off?

racetrack_commons/racetrack_commons/deploy/job_type.py
```diff
@@ -45,7 +48,7 @@ def gather_job_types(
     job_types: dict[str, JobType] = {}
     job_family_versions: dict[str, list[JobType]] = defaultdict(list)
 
-    plugin_results: list[dict[str, tuple[Path, Path]]] = plugin_engine.invoke_plugin_hook(PluginCore.fatman_job_types)
+    plugin_results: list[dict[str, JobTypeImagePaths]] = plugin_engine.invoke_plugin_hook(PluginCore.fatman_job_types)
```
Koshmaar, line 51: Are you sure it's correct change? it's replacing `tuple[Path, Path]` with `list[tuple[Path, Path]]`

iszulcdeepsense: Well, actually it should be `list[tuple[Path, Path]]`, but in the next lines I tried to make it backwards-compatible and also work if `tuple[Path, Path]` was given.

Do you think it should be more strict? (then I'll need to update all the existing job type plugins)

iszulcdeepsense: Or maybe I should use Union type: `list[tuple[Path, Path]] | tuple[Path, Path]`

docs/development/developing-plugins.md
Koshmaar: Good idea :)

racetrack_commons/racetrack_commons/deploy/job_type.py
Koshmaar: Ha, wouldn't have known it could accept both. So yeah, it's good to declare it like that explicitely.

PR #115, opened 2022-12-30
Look around for gorilla/mux alternative

iszulcdeepsense:


pub/health.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"github.com/gin-gonic/gin"
```
Koshmaar, line 7: Nice library you found as alternative to gorilla (which is not maintaned for some time). Has nice design, seems to be actively supported, and is on MIT.

Koshmaar (approved):

PR #135, opened 2023-01-18
Make the admin dashboard a bit easier on the eyes

LookACastle:


dashboard/dashboard/templates/racetrack/administration.html
```diff
@@ -46,10 +46,15 @@ <h5 class="card-title">Infrastructure Targets</h5>
             Available Infrastructure Targets:
             {% if infrastructure_targets|length > 0 %}
                 <ul>
-                {% for infrastructure_target, plugin_manifest in infrastructure_targets.items %}
-                    <li>
-                        <span><b>{{ infrastructure_target }}</b> (by plugin {{ plugin_manifest.name }} {{ plugin_manifest.version }})</span>
-                    </li>
+                {% for infrastructure, instances in infrastructure_instances %}
+                <span>{{ infrastructure }}</span>
+                    <ul>
```
iszulcdeepsense, line 51: I think you missed `<li>` tag as pointed out here https://stackoverflow.com/a/5899394/6772197

> The nested list should be inside a < li > element of the list in which it is nested.

dashboard/dashboard/views.py
```diff
@@ -125,6 +126,15 @@ def view_administration(request):
         context['plugins'] = plugin_client.get_plugins_info()
         context['job_type_versions'] = plugin_client.get_job_type_versions()
         context['infrastructure_targets'] = plugin_client.get_infrastructure_targets()
+
+        #Collect instances running on the same infrastructure
+        _infrastructure_instances: dict = collections.defaultdict(list)
+        for instance, infrastructure in context['infrastructure_targets'].items():
```
iszulcdeepsense, line 132: In fact, there's `infrastructure_name, plugin_manifest` inside this dict and I think it's more accurate to call them that way instead of `instance, infrastructure`

dashboard/dashboard/views.py
```diff
@@ -125,6 +126,15 @@ def view_administration(request):
...
+            _versioned_name: str = infrastructure['name'] + ' (' + infrastructure['version'] + ')'
+            _infrastructure_instances[_versioned_name].append(instance)
+        infrastructure_instances: dict[str, List[str]] = sorted(_infrastructure_instances.items())
```
iszulcdeepsense, line 135: Since you've sorted it, the type is different:
```suggestion
        infrastructure_instances: list[tuple[str, List[str]]] = sorted(_infrastructure_instances.items())
```

dashboard/dashboard/views.py
```diff
@@ -125,6 +126,15 @@ def view_administration(request):
         context['plugins'] = plugin_client.get_plugins_info()
         context['job_type_versions'] = plugin_client.get_job_type_versions()
         context['infrastructure_targets'] = plugin_client.get_infrastructure_targets()
+
+        #Collect instances running on the same infrastructure
+        _infrastructure_instances: dict = collections.defaultdict(list)
+        for instance, infrastructure in context['infrastructure_targets'].items():
+            _versioned_name: str = infrastructure['name'] + ' (' + infrastructure['version'] + ')'
```
iszulcdeepsense, line 133: I believe we can just use the `infrastructure['name']` (or `plugin_manifest['name']`) as a key. If we have 2 infrastructures provided by different versions of plugins, we can group them together.

dashboard/dashboard/templates/racetrack/administration.html
```diff
@@ -66,13 +71,19 @@ <h5 class="card-title">Plugins</h5>
...
+                            {% if plugin.url %}
+                                <span> - <a href="{{ plugin.url }}" target="_blank">{{ plugin.url }}</a></span>
+                            </div>
+                            {% endif %}
+                            {% if user.is_staff %}
+                            <div class="col-md-3 text-end">
+                                <a href="{% url 'dashboard:plugin_config_editor' plugin.name plugin.version %}" class="btn btn-primary">Edit Config</a>
+                                <button type="button" class="btn btn-danger button-delete-plugin" plugin-name="{{ plugin.name }}" plugin-version="{{ plugin.version }}">Delete</button>
+                            </div>
+                        </div>
```
iszulcdeepsense, line 86: I think that `{% endif %}` below should be one line higher, before enclosing `</div>`, to align it with the other side of "if". If user is not staff, the HTML will be broken otherwise.

iszulcdeepsense (changes requested):

dashboard/dashboard/templates/racetrack/administration.html
```diff
@@ -46,9 +46,16 @@ <h5 class="card-title">Infrastructure Targets</h5>
             Available Infrastructure Targets:
             {% if infrastructure_targets|length > 0 %}
                 <ul>
-                {% for infrastructure_target, plugin_manifest in infrastructure_targets.items %}
+                {% for infrastructure, instances in infrastructure_instances %}
+                <span>{{ infrastructure }}</span>
                     <li>
```
iszulcdeepsense, line 51: Sorry to say that, but I guess it should be 
```suggestion
                    <li><span>{{ infrastructure }}</span>
```
cause it still looks wrong in Chrome:
Screenshot from 2023-01-19 13-10-17

iszulcdeepsense (approved):

iszulcdeepsense:
LGTM

PR #137, opened 2023-01-20
Update Copyright notice

iszulcdeepsense:


README.md
```diff
@@ -102,7 +102,7 @@ See the [Documentation](https://theracetrack.github.io/racetrack/) for more deta
 
 ## About
 
-Racetrack is copyright The Danish Business Authority 2022.
+Racetrack is copyright The Danish Business Authority.
```
iszulcdeepsense, line 105: @JosefAssadERST Do I remember correctly that we decided to throw away the year? Does it also apply to License files?

JosefAssadERST: I think so, but [IANAL](https://en.wikipedia.org/wiki/IANAL). Lemme ask an actual lawyer, I'll try to get a quick answer.

JosefAssadERST: The correct copyright notice is:

"This software is copyright Erhvervsstyrelsen"

PR #138, opened 2023-01-20
Show number of total calls in the selected range

iszulcdeepsense:
I hope it makes more sense now:

Screenshot from 2023-01-20 17-13-11

### DIFF
--- a/utils/grafana/dashboards/fatman.json
+++ b/utils/grafana/dashboards/fatman.json
@@ -21,8 +21,8 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "id": 1,
-  "iteration": 1658224133400,
+  "id": 3,
+  "iteration": 1674230877281,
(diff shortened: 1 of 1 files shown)


JosefAssadERST:
~~Ah, is this live anywhere?~~ Ah never mind, I see it's live on C3 dev. Yeah I think this is more useful than what it replaces. @nielshelsoe agree?

nielshelsoe:
Jah I think so. 
Just so I get it. The old counter was just the total of all endpoints called in a Fatman? (E.g. labels,health, perform e.c.t.)

JosefAssadERST:
> Jah I think so. Just so I get it. The old counter was just the total of all endpoints called in a Fatman? (E.g. labels,health, perform e.c.t.)

Yeah.

Old: show total since the Big Bang

New: respect the time range selected on the dashboard up here:

billede

JosefAssadERST (approved):
LGTM

PR #142, opened 2023-01-24
Extend racetrack CLI documentation

iszulcdeepsense:


mkdocs.yml
```diff
@@ -8,6 +8,7 @@ theme:
   name: material
   icon:
     repo: fontawesome/brands/github
+    logo: material/car-sports
```
JosefAssadERST, line 11: This is probably something we should change to something we own sooner rather than later. It's not great to be using something off-the-shelf for a project logo, even if the license permits.

@mariushart can we ask someone in ERST to do this? Do you know anyone that could knock something together on short order?

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
 Racetrack system allows to deploy jobs in a one step.
 It transforms your code to in-operation workloads, e.g. Kubernetes workloads.
 You write some code according to a set of coventions, you include the manifest file which explains the code, 
 and you submit it to Racetrack. A short while after, the service calling your code is in operation.
 
-# Installation
+## Quickstart
+```shell
+# Install racetrack-client
+pip3 install racetrack-client
```
JosefAssadERST, line 12: I have an old habit of using `python3 -m pip install --upgrade racetrack-client`. Some people have problems in their python paths and wonky pip installations, and there might be an old version installed so a forced `--upgrade` can one day be helpful.

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
+
+# Set current remote
```
JosefAssadERST, line 14: Do we explain anywhere what a remote is? We can't assume they have that idea already from git.

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
+This will install `racetrack` CLI tool. Verify installation by running `racetrack` command.
+
+## Usage
+Run `racetrack --help` to see usage.
+
+### Adding a remote
+Assuming your Racetrack server is running on https://racetrack.platform.example.com/lifecycle,
+you can add this remote as an alias:
+```shell
+racetrack set alias my-dev https://racetrack.platform.example.com/lifecycle
```
JosefAssadERST, line 41: `alias` might also need explaining. We can't assume users will hop around between different docs, so it isn't enough to put it in the glossary.

While it might not be possible to achieve 100%, the more we write a doc as if users haven't read the others the better it is (of course, it's a fine balance to avoid actual duplication).

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
 
-# Usage
-Run `racetrack --help` to see usage.
+In case you're going to use a private repository, provide your git credentials so the job can be built from your code:
+```shell
+racetrack set credentials https://github.com/YourUser/YourRepository USERNAME TOKEN
+```
+
+### Installing plugins
+Extend Racetrack's possibilities by installing a bunch of plugins:
```
JosefAssadERST, line 63: Do we want to say that not everyone can do this?

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
+```shell
+racetrack plugin install github.com/TheRacetrack/plugin-python-job-type
```
JosefAssadERST, line 65: It's entirely possible that J. Random User just discovered this project and has only seen this doc, and is executing these steps. Short comments above each explaining why we're installing two and how they're different?

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
+racetrack plugin install github.com/TheRacetrack/plugin-docker-infrastructure
+```
+
+### Deploying a job
+To deploy a job, just run it in the place where `fatman.yaml` is located:
```
JosefAssadERST, line 70: Something like "when your code is ready on your workstation, it's time to deploy it; that means, upload it to RT so it can become a proper running Job" for example.

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
+```shell
+racetrack logs my-job-name
+```
 
-## Deploying
-To deploy a job, just run in the place where `fatman.yaml` is located:
-```bash
-racetrack deploy . --remote https://racetrack.platform.example.com/lifecycle
+### Deleting a job
+Decommission your running job with:
```
JosefAssadERST, line 90: From [George Orwell's Six Rules of Writing](https://infusion.media/blog/george-orwells-six-rules-for-writing/):

    Never use a long word where a short one will do

Nothing wrong with "delete".

racetrack_client/README.md
```diff
@@ -1,30 +1,93 @@
...
+```shell
+racetrack delete my-job-name
```
JosefAssadERST, line 92: If you choose a more whimsical name for this fictional job, it might be easier for the reader to track its progress in this documentation. For example 

```bash
cd src/MuffinDestroyerService
racetrack deploy
```

```bash
racetrack logs MuffinDestroyer
```

```bash
racetrack delete MuffinDestroyer
```

mkdocs.yml
iszulcdeepsense: Right. Just so you know, without this line, the default comes in and we were using default mkdocs's logo so far
Screenshot from 2023-01-27 13-02-24
so I thought it's better to set it to something more fitting.

As you said material icons are under permissive [Apache 2.0 License](https://github.com/google/material-design-icons/blob/master/LICENSE)

racetrack_client/README.md
iszulcdeepsense: Good idea. I've explained what alias and remote are:

> Alias is a short, friendly name for the URL of your Racetrack server, which is also known as "remote".
From now on, you can refer to your remote with an alias.

mkdocs.yml
JosefAssadERST: Agree it's better than the default. :) It's still valuable having our own. It gives us something to defend if RT becomes too successful.

PR #149, opened 2023-01-27
Set User-agent in all CLI requests

iszulcdeepsense:


racetrack_client/racetrack_client/utils/request.py
```diff
@@ -190,8 +192,18 @@ def _make_request(cls,
...
+            req.add_header('User-Agent', 'request')
+
         kwargs['context'] = cls._get_ssl_context()
 
+        if DEBUG_MODE:
+            http_handler = request.HTTPHandler(debuglevel=2)
+            https_handler = request.HTTPSHandler(context=kwargs['context'], debuglevel=2)
+            opener = request.build_opener(http_handler, https_handler)
+            request.install_opener(opener)
+            del kwargs['context']
```
Koshmaar, line 205: Out of curiosity - why do you delete context only in debug mode? Will conflict with something else?

Koshmaar (approved):

racetrack_client/racetrack_client/utils/request.py
iszulcdeepsense: Good question. I spent a lot of time debugging this. Here's the implenetation of urllib's urlopen:
```python
def urlopen(url, data=None, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
            *, cafile=None, capath=None, cadefault=False, context=None):
    global _opener
    if cafile or capath or cadefault:
        ...
    elif context:
        https_handler = HTTPSHandler(context=context)
        opener = build_opener(https_handler)
    elif _opener is None:
        _opener = opener = build_opener()
    else:
        opener = _opener
    return opener.open(url, data, timeout)
```
and that means if you pass the `context` argument, the "opener" will be rebuilt, not giving you a chance to turn the debug mode on.

iszulcdeepsense: I'll add the comment there

PR #151, opened 2023-01-31
More granular deployment phases

iszulcdeepsense:


docker-compose.yaml
```diff
@@ -81,6 +81,7 @@ services:
       DOCKER_TLS_CERTDIR: ''
       DOCKER_HOST: tcp://docker-engine:2375
       CONFIG_FILE: /etc/racetrack/image-builder/compose.yaml
+      LIFECYCLE_TOKEN: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzZWVkIjoiZGVhNzYzMDAtN2JhYy00ODRjLTkyOTYtNWQxMGNkOTM3YTU4Iiwic3ViamVjdCI6ImltYWdlLWJ1aWxkZXIiLCJzdWJqZWN0X3R5cGUiOiJpbnRlcm5hbCIsInNjb3BlcyI6WyJmdWxsX2FjY2VzcyJdfQ.ND3wDeK58L5T1jIYcuArQ5O3M0Ez3_pCAEi5NXD_hLY
```
Koshmaar, line 84: Nitpick: changing name to LIFECYCLE_AUTH_TOKEN would make it more clear what is this token for. I had doubts what it is for, untill I saw rest of code.

image_builder/image_builder/docker/builder.py
```diff
@@ -55,38 +56,48 @@ def build(
...
+                with wrap_context(f'building base image {progress}'):
+                    update_deployment_phase(config, deployment_id, f'building image {build_progress}')
+                    base_image = _build_base_image(
+                        config, job_type, image_index, deployment_id, metric_labels, build_progress,
+                    )
+
+                build_progress = f'({image_index*2+2}/{images_num*2})'
+                template_path = job_type.template_paths[image_index]
+                if template_path is None:
+                    assert manifest.docker and manifest.docker.dockerfile_path, 'User-module Dockerfile manifest.docker.dockerfile_path is expected'
```
Koshmaar, line 71: This build function is one hell of a function :] long, with loop, multiple if statements, exception handling... it has high cognitive complexity. Suggestion here is to split it to subfunctions.

docker-compose.yaml
Koshmaar: Ah, I see now it's JWT - interesting, didn't notice when we started using them. One more thing, so how does it work in production, we have to pass to image builder static token of some admin user?

Koshmaar (approved):

docker-compose.yaml
iszulcdeepsense: Sure, I'll rename it.

You can decode this token at https://jwt.io/ and see what's inside. These auth tokens can have different `subject_type`, like `user` or `internal`. In this case it's "internal" token, so it doesn't belong to any user, we can call it a service account.

To generate such token, you can call `AUTH_KEY=s3cr3t python -m lifecycle generate-auth image-builder`. For more details, start from [here](https://github.com/TheRacetrack/racetrack/blob/master/lifecycle/lifecycle/main.py#L25)

In production there is a different `AUTH_KEY`, so these dev tokens won't work due to invalid signature.

PR #160, opened 2023-02-03
Host local Grafana

iszulcdeepsense:


utils/prometheus/prometheus.yaml
```diff
@@ -31,11 +31,10 @@ scrape_configs:
...
     metrics_path: /metrics
-    dns_sd_configs:
-      - names: 
-          - 'fatman-adder-v-0-0-1'
-        type: A
-        port: 7000
+    file_sd_configs:
+      - files:
+        - '/mnt/plugins/metrics/sd_config_jobs.yaml'
+        refresh_interval: 1m
```
Koshmaar, line 40: [nitpick]This single entry will serve all jobs that are deployed, right? Then naming it `racetrack-jobs` would make it more clear what it's doing. [/nitpick]

lifecycle/lifecycle/supervisor/metrics.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    metrics_dir = Path(config.plugins_dir) / 'metrics'
+    ensure_dir_exists(metrics_dir)
+    service_discovery_file: Path = metrics_dir / 'sd_config_jobs.yaml'
+
+    file_content: list[str] = []
+    with wrap_context('populating Prometheus configuration'):
+        for job in list_fatmen_registry(config):
+            if job.status != FatmanStatus.LOST.value:
+                file_content.append(_build_job_config(job))
```
Koshmaar, line 26: This means that when Fatman starts having problems, the metrics for it won't be collected for the nearest minute (or longer). I wonder how Prometheus reacts to such disappearing jobs - my assumption is it won't forget them, right? And if job reappears, it will pick it up, and collect metrics to the same job. It makes sense that's how it should work, but would be good to confirm.

Koshmaar (approved):

utils/prometheus/prometheus.yaml
iszulcdeepsense: Right, renaming makes sense

lifecycle/lifecycle/supervisor/metrics.py
iszulcdeepsense: Yep, if a job disappears, Prometheus still keeps track of it, job has just null values then. When it appears again, it picks it up.

"when Fatman starts having problems" it usually has "ERROR" status and notice that Prometheus will still try to scrape from it, that's intended.  I only excluded "LOST" status as it usually means that the cluster was wiped out and there's no need to bother Prometheus.

PR #166, opened 2023-02-10
Watch database connection

iszulcdeepsense:


lifecycle/lifecycle/server/metrics.py
```diff
@@ -24,8 +28,26 @@ def collect(self):
...
     except OperationalError:
         return False
+
+
+def is_port_open(ip: str, port: int) -> bool:
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(3)
+    try:
+        s.connect((ip, port))
+        s.shutdown(2)
```
LookACastle, line 50: `shutdown(2)` is a little opaque if you aren't aware of what that means already. I prefer the following
```suggestion
        s.shutdown(socket.SHUT_RDWR)
```

iszulcdeepsense: Good point, thanks.

LookACastle (approved):
LGTM

PR #184, opened 2023-02-21
Set mode 600 on racetrack config file

iszulcdeepsense:
It sets mode 600 each time the config is rewritten. So the existing files will be fixed once any configuration change happens, eg. `racetrack set remote ...`

@blackplague  I can't set you as a reviewer, but please take a look.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Racetrack client keeps the local config file `~/.racetrack/config.yaml` with mode `600`
+  (not readable by others) and warns you if this file has insecure permissions
+  since it may contain credentials.
 
 ## [2.9.1] - 2023-02-20
 ### Changed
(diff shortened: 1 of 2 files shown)


JosefAssadERST:
I feel like it's a little imperialist for RT to actively change file mode itself. ssh doesn't try to change the permissions on `~/.ssh` or its contents, it just complains loudly and refuses to run.

I think that's a more correct behavior. It makes the file mode requirement more explicit, and on the off chance there's some other software out there that's also manipulating this file or directory, it prevents clobbering. It's just generally polite to advise than to intervene.

blackplague:
I agree with @JosefAssadERST in the sense that I think RT should create the original config file with permissions 0600 and I am fine with it validating and throwing hissy fits in case a user decides that it would be amazing to have 0777 permissions on a file containing access tokens and credentials.
Regarding .ssh, I have never changed the permission on the .ssh directory and it comes out of the box with user - read, write and execute at least on my GNU/Linux distro.

iszulcdeepsense:
Sure, how about showing a warning message instead of refusing to run?
If we go with the second, I'm afraid we'll piss off all the active users.

JosefAssadERST:
If someone has 0777 I think they'll be more ashamed than pissed off... @blackplague do you think we have any sinners like that?

blackplague:
No, we don't have any of the 0777 sinners, but we have a bunch of 0664 sinners, which again is too lax. Only two people have their config.yaml restricted, one of them being me.

JosefAssadERST:
I'm ok with @iszulcdeepsense 's idea of just warning, even though my personal preference is to refuse to run. That file can have actual secrets and I don't mind educating users about this kind of permission.

iszulcdeepsense:
> Just a note, this is not a democracy. @iszulcdeepsense decides.

image

iszulcdeepsense:
Remarks resolved. Please have a look again.

blackplague (approved):
I think it look good. I did consider whether we should check for whether the file is also writable by others. But hopefully that is not necessary.

PR #185, opened 2023-02-23
Add endpoint for regenerating individual user tokens.

### LINKED ISSUE #139: add ability to regenerate token
This:

image

LookACastle:
Closes #139.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,9 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Racetrack client keeps the local config file `~/.racetrack/config.yaml` with mode `600`
   (not readable by others) and warns you if this file has insecure permissions
   since it may contain credentials.
-- Endpoint `/api/v1/auth/token/user/regenerate`'s previous functionality moved to `/api/v1/auth/token/user/all/regenerate 
 ### Added
-- Endpoint `/api/v1/auth/token/user/regenerate` added to regenerate tokens on a per-user basis.
+- Button added to the profile page that lets you regenerate your authentication token.
 
 ## [2.9.1] - 2023-02-20
 ### Changed
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Racetrack client keeps the local config file `~/.racetrack/config.yaml` with mode `600`
   (not readable by others) and warns you if this file has insecure permissions
   since it may contain credentials.
+- Endpoint `/api/v1/auth/token/user/regenerate`'s previous functionality moved to `/api/v1/auth/token/user/all/regenerate 
 ### Added
-- Endpoint `/api/v1/auth/token/user/<username>/regenerate` added to regenerate tokens on a per-user basis.
+- Endpoint `/api/v1/auth/token/user/regenerate` added to regenerate tokens on a per-user basis.
 
 ## [2.9.1] - 2023-02-20
 ### Changed
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Racetrack client keeps the local config file `~/.racetrack/config.yaml` with mode `600`
   (not readable by others) and warns you if this file has insecure permissions
   since it may contain credentials.
+### Added
+- Endpoint `/api/v1/auth/token/user/<username>/regenerate` added to regenerate tokens on a per-user basis.
 
 ## [2.9.1] - 2023-02-20
 ### Changed
--- a/lifecycle/lifecycle/auth/subject.py
+++ b/lifecycle/lifecycle/auth/subject.py
@@ -135,6 +135,13 @@ def _get_subject_name_from_auth_subject(auth_subject: models.AuthSubject) -> str
     else:
         raise ValueError("Unknown auth_subject type")
 
+def regenerate_user_token(uesrname: str):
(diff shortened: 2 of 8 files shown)


lifecycle/lifecycle/endpoints/auth.py
```diff
@@ -80,6 +80,12 @@ def _get_esc_token(esc_id: str, request: Request):
         auth_subject = get_auth_subject_by_esc(esc_model)
         return {'token': auth_subject.token}
 
+    @api.post('/auth/token/user/{username}/regenerate')
+    def _generate_tokens_for_user(request: Request, username: str):
+        """Generate new token for a User"""
+        check_auth(request, scope=AuthScope.CALL_ADMIN_API)
```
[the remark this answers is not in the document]
iszulcdeepsense: Also I think we don't need `username` parameter here, we don't want to allow to reset someone else's token. We can take that information from the current user calling this endpoint
```suggestion
        auth_subject = check_auth(request, subject_types=[AuthSubjectType.USER])
        regenerate_user_token(auth_subject)
```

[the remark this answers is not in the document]
LookACastle: Would it not make sense for an admin to be able to reset a misbehaving token without the owner of said token present?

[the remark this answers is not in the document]
iszulcdeepsense: True, but admin can already do that on the Django Administration Panel (see "Generate token" button - it makes a new one)
Screenshot from 2023-02-24 09-53-39
I think this issue is rather about giving that possibility to the regular users.

lifecycle/lifecycle/endpoints/auth.py
```diff
@@ -80,6 +80,12 @@ def _get_esc_token(esc_id: str, request: Request):
         auth_subject = get_auth_subject_by_esc(esc_model)
         return {'token': auth_subject.token}
 
+    @api.post('/auth/token/user/{username}/regenerate')
+    def _generate_tokens_for_user(request: Request, username: str):
```
iszulcdeepsense, line 84: Since the argument `username` is not used, it can be deleted.

This might be misinterpreted by someone using this API. 
Imagine the Administrator who finds this endpoint on a Swagger page. He thinks it would be great to regenerate someone else's token, but to his surprise, he ends up with its own token being regenerated.

lifecycle/lifecycle/auth/subject.py
```diff
@@ -135,6 +136,9 @@ def _get_subject_name_from_auth_subject(auth_subject: models.AuthSubject) -> str
     else:
         raise ValueError("Unknown auth_subject type")
 
+def regenerate_specific_user_token(user: User):
+    regenerate_auth_token(user)
```
iszulcdeepsense, line 140: What you're actually passing to this function is `auth_subject: models.AuthSubject`. (Auth Subject can be either User, Job or External Consumer)
Also `regenerate_auth_token` expects auth_subject, and that's ok, so just the type annotations are wrong.

```suggestion
def regenerate_specific_user_token(auth_subject: models.AuthSubject):
    regenerate_auth_token(auth_subject)
```

lifecycle/lifecycle/endpoints/auth.py
iszulcdeepsense: Ah, I see then it collides with already existing one `/auth/token/user/regenerate`, I think it's more reasonable to rename the latter to `/auth/token/user/all/regenerate` and then we can reclaim `/auth/token/user/regenerate`.

iszulcdeepsense (commented):
Looks like regenerating a token causes troubles with a current session, since the token is in use when you're logged in. You can observe it when opening a Jobs tab right after regenerating a token.

Here's the fix. Sorry for the .patch.zip ,but I can't commit to your repo and Github forbids uploading .patch files. You can apply it with `git am 0001-Regenerate-user-session.patch.zip`
[0001-Regenerate-user-session.patch.zip](https://github.com/TheRacetrack/racetrack/files/10848906/0001-Regenerate-user-session.patch.zip)

lifecycle/lifecycle/endpoints/auth.py
LookACastle: Yeah, I kept the naming scheme for exactly the reason you suspect. Are we good to break API like that, or should I change this pull request so it goes into some "future major version" branch rather than master?

LookACastle: There's an alternative for just adding a check that a user only ever gets to call the endpoint with their own name in it.

iszulcdeepsense: I'm completely fine to change the path of `_generate_tokens_for_all_users` endpoint. You can notice that it's not being used by the code. It's an administrative endpoint to be clicked by admin from a Swagger page.

docs/CHANGELOG.md
```diff
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Racetrack client keeps the local config file `~/.racetrack/config.yaml` with mode `600`
   (not readable by others) and warns you if this file has insecure permissions
   since it may contain credentials.
+- Endpoint `/api/v1/auth/token/user/regenerate`'s previous functionality moved to `/api/v1/auth/token/user/all/regenerate 
 ### Added
-- Endpoint `/api/v1/auth/token/user/<username>/regenerate` added to regenerate tokens on a per-user basis.
+- Endpoint `/api/v1/auth/token/user/regenerate` added to regenerate tokens on a per-user basis.
```
iszulcdeepsense, line 14: Nitpicking: In my opinion there's no need to be so specific. Notice that we just want to keep **user-facing** changes in here. So probably "There is a new button to regenerate your auth token" will be interesting enough for our "users".

iszulcdeepsense (approved):

PR #187, opened 2023-02-27
Make Grafana dashboards more rational

iszulcdeepsense:


Koshmaar:
Unfortunately I'm not grafana expert, so it's hard to review it through source diff. But probably you have it running somewhere where I could look? Or at least few screenshots :)

iszulcdeepsense:
> Unfortunately I'm not grafana expert, so can't review it solely looking at source diff. But probably you have it running somewhere where I could look? Or at least few screenshots :)

Indeed, I didn't give you a chance to test it. I've PM'ed you a link.
At least granularity looks better now:
Screenshot from 2023-02-28 16-37-49

iszulcdeepsense:
Hang on, I see there is a problem with the jobs having more than 1 replica. Prometheus scrapes metrics indirectly, through a loadbalanced service, thus getting different metrics each time.

I'm turning it into a draft , will get back to you once it's fixed.

iszulcdeepsense:
Now it's finally solved. There was a problem with replicas. Prometheus wrongly thought there is a change cause the metric values were constantly switching. Now it only show the traffic where it really occurs:

Screenshot from 2023-03-02 10-20-28

racetrack_commons/racetrack_commons/entities/dto.py
```diff
@@ -38,6 +39,7 @@ class JobDto(BaseModel, arbitrary_types_allowed=True):
     deployed_by: Optional[str] = None
     last_call_time: Optional[int] = None
     infrastructure_target: Optional[str] = None
+    replica_internal_names: list[str] = []
```
Koshmaar, line 42: The purpose of this field is quite cryptic, adding comment what is its purpose would be useful (without knowing about this MR, I would have thought it's for some internal scaled pod handling, for sure not for something like grafana/prometheus).

Koshmaar (approved):

PR #192, opened 2023-03-03
Upgrade dependencies

iszulcdeepsense:


racetrack_client/racetrack_client/utils/quantity.py
```diff
@@ -81,3 +81,19 @@ def __truediv__(self, other) -> 'Quantity':
...
+    def validate(cls, v):
+        if v is None:
+            return None
+        return Quantity(str(v))
+
+    @classmethod
+    def __modify_schema__(cls, field_schema):
+        field_schema.update(
+            examples=["128974848", "129M", "128974848000m", "123Mi", "0.129G"],
+        )
```
Koshmaar, line 99: Hmm, are those three functions needed only to get better Pydantic compatibility? And they're the only methods added for Pydantic? If so, I propose to mark them as such with comment, so coders will understand where to look for more documentation (or maybe only I don't know Pydantic).

Koshmaar (approved):

racetrack_client/racetrack_client/utils/quantity.py
iszulcdeepsense: Yes, without these lines, new Pydantic fails to operate. And yes, I'll add comments that it's intended for pydantic.

Issue #163, opened 2023-02-07
Abandon adopting orphaned jobs


iszulcdeepsense:
Sometimes there's a chance to have empty manifests in the Jobs of Racetrack.
It happens when a job was created by other Racetrack instance and then it's found by another one and adopted, then we don't have a manifest in our database, cause we see only a name of a job.

Racetrack discovers wild jobs by labels. Infrastructure targets implement how to list jobs, eg. k8s infrastructure makes `kubectl get pods` with label filter. The result is compared with the database state and it can be used not only to check what's missing, but also to find out what's extra.

Racetrack shouldn't adopt "orphans" any longer. We gain nothing significant from it.

JosefAssadERST:
I agree with the idea. Should there be admin facilities to reap them?

iszulcdeepsense:
> I agree with the idea. Should there be admin facilities to reap them?

I think we aren't gonna need it. Can rethink it once the need arises.

JosefAssadERST:
For these orphans, don't we have some sort of responsibility towards the owners of our infrastructure targets? On a bad day we could end up with a lot of orphans which are then taking up resources for no gain. That's why I was thinking of reaping.

The infrastructure admin can't just go do the reaping themselves since they don't know which of the pods with Racetrack labels are orphans and which are not.

iszulcdeepsense:
Good point, it would be nice to do the reaping, but only Racetrack knows what to reap. 
I wonder how to notify the admin about such resources to spare. Here's the idea:
Lifecycle could print warning logs about the orphaned jobs (along with more details like pod names) and expose a Prometheus metrics with the amount of the orphaned jobs. Then, you could set up an alert in Grafana whenever number of orphans is greater than 0 and that's how Admin would be notified. He peeks into the logs and does the reaping then.

I'm not super crazy about user's convenience in this case, cause it's very unlikely to happen.

JosefAssadERST:
That's pretty minimal, which I think is good. If we need to make it more complicated we can later.

Issue #134, opened 2023-01-18
Spurious "Lost" status


iszulcdeepsense:
Sometimes it happens that Racetrack wrongly claims "Lost" status of the Job, while it is really running.

I think there's a chance for a race condition when the background thread finds the newly created Pod, while it's in the middle of deployment and it's not ready yet, so it claims it to be Lost. Not sure if this still occurs, need to explore that.

 Refreshing the status by `lifecycle-supervisor` happens in background every minute, so there's a chance that your Job is really running, while the status is still Lost and now you need to wait a minute until it's RUNNING again.

PR #193, opened 2023-03-06
Abandon adopting orphaned jobs

iszulcdeepsense:
Closes https://github.com/TheRacetrack/racetrack/issues/163
Closes https://github.com/TheRacetrack/racetrack/issues/134

Summary of changes:
- Racetrack doesn't adopt "orphaned" jobs any longer. Lifecycle-supervisor just prints the warning log about the orphaned job.
- a new Prometheus metric is exposed - the number of jobs by status.
- This metric is the first metric of lifecycle-supervisor, so it was added to prometheus config.
- In Grafana dashboards, there's a new panel monitoring the number of jobs with the particular status.
  
Screenshot from 2023-03-07 13-24-30


lifecycle/lifecycle/job/registry.py
```diff
@@ -124,7 +126,7 @@ def sync_registry_jobs(config: Config, plugin_engine: PluginEngine):
         cluster_jobs = _generate_job_map(list_cluster_jobs(config, plugin_engine))
         registry_jobs = _generate_job_map(list_job_registry(config))
         logger.debug(f'Found {len(cluster_jobs)} jobs in the cluster, {len(registry_jobs)} in the database')
-        job_status_count: Dict[str, int] = defaultdict(int)
+        job_status_count: dict[str, int] = defaultdict(int)
```
mpolomdeepsense, line 129: Why this change? It doesn't change anything and you have to write an additional import for it to work, and since you are still importing `Dict` from `typing` (used at the bottom of this file) this change seems even more meaningless.

iszulcdeepsense: Python 3.9 introduced this https://peps.python.org/pep-0585/. Starting from Python 3.9, using the generic types is now preferred rather than their `typing` counterpart. 
So I keep replacing it once I see it somewhere around.

Koshmaar (approved):
LGTM. Just one thing that could be improved, is adding some developer doc about the design change "RT doesn't adopt orphaned jobs, because (...), we do it differently (...)". Not sure if you're documenting this kind of logic, but I imagine that if I was a new developer starting in this project, getting to know such details of logic would involve long grepping through code.

lifecycle/lifecycle/job/registry.py
iszulcdeepsense: As you said, there is still `typing.Dict` import, so perhaps I should replace all occurrences in the whole file to get rid of it.

Koshmaar:
Yeah, agree that making such documentation is energy costly, and we should be conserving energy. Also docs get outdated faster than comments, so what you propose here is ok for me. But if I may clarify something, I didn't mean putting details of all functions in the docs, just high level overview, ie. https://rep.erst.dk/git/kubernilla/recordkeeper/-/tree/master/graph_builder#implementation-details

PR #196, opened 2023-03-09
Wildcards in job type versions

iszulcdeepsense:


lifecycle/lifecycle/django/app/settings.py
```diff
@@ -106,7 +111,7 @@ def is_env_flag_enabled(flag_name: str, default: str) -> bool:
 
 USE_I18N = True
 
-USE_L10N = True
+USE_L10N = False
```
Koshmaar, line 114: This is interesting - why do you disable localization? is it needed for the wildcard changes, or by the way? I see in Django 5 this env is deprecated, but you're still on 4..

racetrack_client/racetrack_client/utils/semver.py
```diff
@@ -44,18 +44,26 @@ def find_latest_stable(objects: List[T], key: Callable[[T], str]) -> Optional[T]
...
+        objects: List[T],
+        key: Callable[[T], str],
+        only_stable: bool = True,
+    ) -> Optional[T]:
         """
         Find object with the latest (highest), stable (non-dev) version matching pattern
         :param pattern: version pattern containing "x" wildcards, eg. "1.2.x", "2.x"
         :param objects: List of objects to compare
         :param key: Callable to extract version from an object
+        :param only_stable: whether to filter only stable versions
```
Koshmaar, line 58: You could add here not what does it mean that version is stable.

racetrack_client/racetrack_client/utils/semver.py
```diff
@@ -125,26 +133,59 @@ def __hash__(self):
...
-        match = SemanticVersionPattern.valid_pattern_regex.fullmatch(pattern)
+    def is_x_wildcard_pattern(pattern: str) -> bool:
+        """Check whether it's a valid version pattern containing "x" wildcards"""
+        match = SemanticVersionPattern.x_pattern_regex.fullmatch(pattern)
         if not match:
             return False
         return 'x' in pattern
+
+    @staticmethod
+    def is_asterisk_wildcard_pattern(pattern: str) -> bool:
```
Koshmaar, line 186: Nitpick: maybe removing `wildcard` from the name of this function, and the one above, would make them less vocal, simpler and easier to understand. For me the word `wildcard` is almost synonymous with `?*` chars. So seeing `is_x_wildcard_pattern` I start thinking that this function looks for `x` or `*` chars. Maybe it's just me.. for consideration.

Koshmaar (approved):
A lot of regexes, but you secured yourself with battery of tests, so good PR.

lifecycle/lifecycle/django/app/settings.py
iszulcdeepsense: I changed it by the way when I noticed the datetime format is wrong. ISO format was supposed to be displayed in Django, but got this:
Screenshot from 2023-03-14 18-29-16
Sorry for putting it here, I think it's too tiny to make an individual PR.

PR #198, opened 2023-03-10
Setup CSRF Trusted Origins

iszulcdeepsense:


Koshmaar (approved):
Approving, with asterisk that those changes depend on env name which is discussed in other MR.

PR #201, opened 2023-03-13
Reset cookie after regenerating a token

iszulcdeepsense:


dashboard/dashboard/middleware.py
```diff
@@ -40,28 +42,40 @@ def __call__(self, request):
...
+
+    # Domain defines the host to which the cookie will be sent.
+    # Empty domain means that cookie will be created for current host, not including subdomains.
+    # if specified, then subdomains are included.
+    response.set_cookie(
+        key=RT_AUTH_HEADER,
+        value=value,
+        path='/',
+        domain=_get_cookie_domain(),
+        max_age=60 * 60 * 24 * 365,  # one year of cookie lifetime (in seconds), also makes cookie persist between browser restarts
```
LookACastle, line 70: Is there a reason this lasts an entire year? I don't think it's an issue, it just seems like quite a long time intuitively.

LookACastle (approved):
Looks good to merge

dashboard/dashboard/middleware.py
JosefAssadERST: Is this cookie age? FWIW I routinely set that to ten years in my own development projects. I think it's pretty common practise.

iszulcdeepsense: Yes, it's the cookie age. I also think a year is quite a long time. Do we want to decrease it to a month?

LookACastle: I think a month makes more sense, but just chose whatever makes sense to you. The exact value isn't super important, a year just seemed odd.

iszulcdeepsense: So I'm gonna change it to month. @JosefAssadERST  Any objections?

JosefAssadERST: No objections, but I'm curious. What ia gained by going from 1 year to 1 month?

iszulcdeepsense: I guess that's just a good practice in terms of security. If an attacker gains access to a user's computer, they can steal the cookies and impersonate the user. Setting a short lifetime for cookies limits the time frame in which an attacker can use stolen cookies, reducing the risk of a successful attack.

Plus, we'll make sure users remember their password :wink:

PR #208, opened 2023-03-15
Reduce number of requests made by pub

iszulcdeepsense:
PUB was making 3 requests per each job call:
- get job details (internal URL, exact version, etc.)
- get job's public endpoints
- check if user is allowed to speak to a job

Now it's all done with just 1 request to Lifecycle.

Fun fact: this PR removes more lines than it adds.

### DIFF
--- a/lifecycle/lifecycle/endpoints/auth.py
+++ b/lifecycle/lifecycle/endpoints/auth.py
@@ -145,6 +145,10 @@ def _normalize_endpoint_path(endpoint: str) -> str:
 
 
 def _authorize_job_caller(job_model: models.Job, endpoint: str, request: Request):
+    """
+    Check if auth subject (read from request header's token) is allowed to call an endpoint of a job.
+    In case it's not allowed, raise UnauthorizedError.
+    """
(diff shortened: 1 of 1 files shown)


lifecycle/lifecycle/endpoints/auth.py
```diff
@@ -104,3 +135,30 @@ def _generate_tokens_for_all_escs(request: Request):
...
+
+def _normalize_endpoint_path(endpoint: str) -> str:
+    if endpoint.endswith('/'):
+        endpoint = endpoint[:-1]
+    if not endpoint.startswith('/'):
+        endpoint = '/' + endpoint
+    return endpoint
+
+
+def _authorize_job_caller(job_model: models.Job, endpoint: str, request: Request):
```
Koshmaar, line 147: Nitpick: looking at just the function definition in the MR, I had problem understanding how it accepts and rejects the requests, I had to open IDE and lookup the internal functions. So I think adding short comment about the thrown exception in case of rejection etc could be useful.

Koshmaar (approved):
Great stuff :)

PR #214, opened 2023-03-17
Add job type column to racetrack client list command

iszulcdeepsense:
### Added
- New column in the *list* command of the Racetrack client displays a job type.
  Try it out by running `racetrack list -c job_type`.

### Changed
- The *list* command of the Racetrack client drops the fancy formatting and *INFO*/*DEBUG* logs
  when being piped into another command (ie. not connected to a terminal/tty device). Try `racetrack list | cat`.


zcg909 (approved):
I have verified that this works nicely

PR #220, opened 2023-03-22
Add "perform" command to call a job with CLI

iszulcdeepsense:
- New command `racetrack call NAME ENDPOINT PAYLOAD [--version VERSION] [--remote REMOTE] [--curl]`
  allows you to call an endpoint of a job.
  Provide the name of the job and the payload of the request in JSON or YAML format,
  for example `racetrack call adder /api/v1/perform '{"numbers": [40,2]}'`.
  Use `--curl` flag, if you want to generate a curl query instead of calling the job.
  Check out `racetrack call --help` for more details.
- Name of the job can be autocompleted by hitting `Tab` while typing a CLI command.
  Remember to run `racetrack --install-completion` beforehand.
  Under the hood, it fetches the available jobs from the current remote.


LookACastle (approved):
LGTM

PR #222, opened 2023-03-23
Fill in the missing installation guide for non local clusters

iszulcdeepsense:


docs/deployment/k8s-installation.md
```diff
@@ -0,0 +1,280 @@
...
+Use one of these tools to inspect your cluster resources:
+
+- `kubectl`
+- Cloud Console
+- [Kubernetes Dashboard](#deploy-kubernetes-dashboard)
+- [k9s](https://github.com/derailed/k9s)
+
+- Check what resources you're actually trying to deploy with `kubectl kustomize kustomize/aks`
+
+### Deploy Kubernetes Dashboard
```
LookACastle, line 219: Is this the purview of racetrack? That is, can we not expect that if someone is deploying on kubernetes and want a ui troubleshooting thing of some sort, they are going to do that for themselves?

This seems out of scope for the guide to me.

kustomize/aks/dashboard.yaml
```diff
@@ -0,0 +1,86 @@
...
+            - name: EXTERNAL_PUB_URL
+              value: 'http://1.1.1.1/pub'
+            - name: DJANGO_DEBUG
+              value: 'true'
+            - name: AUTH_REQUIRED
+              value: 'true'
+            - name: DASHBOARD_PORT
+              value: '7003'
+            - name: LIFECYCLE_AUTH_TOKEN
+              value: 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzZWVkIjoiMjIzYThhM2ItNzBkOC00YzA4LWI4YmYtNzlkNDA3MDI3ZGE2Iiwic3ViamVjdCI6ImRhc2hib2FyZCIsInN1YmplY3RfdHlwZSI6ImludGVybmFsIiwic2NvcGVzIjpbImZ1bGxfYWNjZXNzIl19.CAww41hboBPHNflJxKBNRnO-Eo9Y_EKsfU66mK3kdk4'
```
LookACastle, line 51: It seems like a mistake to have a token stored like this. Even if it's just the default value, perhaps that'd be okay to ask the users to fill out for themselves.

kustomize/aks/pub.yaml
```diff
@@ -0,0 +1,71 @@
...
+            - name: PUB_PORT
+              value: '7005'
+            - name: LIFECYCLE_URL
+              value: 'http://lifecycle:7002/lifecycle'
+            - name: AUTH_REQUIRED
+              value: 'true'
+            - name: AUTH_DEBUG
+              value: 'true'
+            - name: LIFECYCLE_AUTH_TOKEN
+              value: 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzZWVkIjoiY2E3ZWNmODAtZDAxZS00ZjA5LWI4ZjQtMTZiNTMxYjhhOTNhIiwic3ViamVjdCI6InB1YiIsInN1YmplY3RfdHlwZSI6ImludGVybmFsIiwic2NvcGVzIjpbImZ1bGxfYWNjZXNzIl19.Xyg1lwQG4oiA65y-KAkdmCFjnT7BZkfmm66DDGOVZu4'
```
LookACastle, line 42: Ditto the other token comments.

LookACastle (commented):
A more general comment, why `kustomize/aks`? This is not unique to aks, is it? If so, wouldn't it make a lot more sense to have `kustomize/k8s` or something along those lines?

iszulcdeepsense:
@LookACastle 
> A more general comment, why `kustomize/aks`? This is not unique to aks, is it? If so, wouldn't it make a lot more sense to have `kustomize/k8s` or something along those lines?

"k8s" sounds tempting. However, we got "kind" in the same folder, which is also kubernetes instance. So I would rather call it maybe "remote", "external" or "cloud"?
Plus, unfortunately, it turns out that this is not so generic as it's expected to be. Azure requires setting up this line in ingress-controller:
`service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path: "/healthz"`
Note, **azure** keyword, that's why it's more associated with "aks".

docs/deployment/k8s-installation.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Prepare Docker Registry
+Racetrack needs a Docker registry to store the images of the jobs.
+We need to instruct Kubernetes to pull images from there.
+
+Let's assume we have a Docker registry at `ghcr.io/theracetrack/racetrack/` with
+`racetrack-registry` user and `READ_REGISTRY_TOKEN` and `WRITE_REGISTRY_TOKEN`
+tokens for reading and writing images respectively.
+
+Create a secret for the registry in **kustomize/aks/docker-registry-secret.yaml**.
```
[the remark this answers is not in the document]
iszulcdeepsense: It's provided indeed, but it's supposed to be overwritten with user's tokens. I'll make a note.

docs/deployment/k8s-installation.md
```diff
@@ -0,0 +1,280 @@
...
+spec:
+  type: LoadBalancer
+...
+```
+
+Now, let's make Racetrack aware of this IP address.
+If you don't know the exact IP address of the LoadBalancer, you can skip this step for now.
+It will be assigned after the first deployment, then you can get back to this step, set the IP address and apply it again.
+
+Fill in the public IP in the following places:
```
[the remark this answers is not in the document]
iszulcdeepsense: Right.
By the way, is there a way to somehow use env var automatically by kustomize and fill it in the template? (I think what we'd like to do is called "helm charts", but maybe kustomize has a feature I don't know about)

docs/deployment/k8s-installation.md
iszulcdeepsense: You're right. It was more like the notes for myself, I'll cut it out.

kustomize/aks/dashboard.yaml
iszulcdeepsense: Actually, this token is not a user's token, but the one for internal communication of components and it's more complicated to generate this one. Plus, notice that each of them is different. There's 3 of them: one for each dashboard, image-builder and pub.
They can be generated offline, before launching the Racetrack, but the process is a bit complicated. That's why I omitted it and went with the default ones.
Do you think we should put that as well in this documentation?

docs/deployment/k8s-installation.md
```diff
@@ -0,0 +1,280 @@
...
+kubectl config get-contexts
+kubectl config use-context aks-racetrack # k8s context is `aks-racetrack` in this tutorial
+kubectl config set-context --current --namespace=racetrack
+```
+
+## Prepare Docker Registry
+Racetrack needs a Docker registry to store the images of the jobs.
+We need to instruct Kubernetes to pull images from there.
+
+Let's assume we have a Docker registry at `ghcr.io/theracetrack/racetrack/` with
```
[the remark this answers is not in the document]
iszulcdeepsense: I'm afraid Racetrack still needs a Docker registry. While we could use the `ghcr.io` and we could push the **base** images beforehand so that Racetrack sees they're already there and that's okay. However, in the next step Racetrack builds the **job** image which is always unique and has a unique tag composed of a current timestamp. Thus, it always needs the registry to deploy jobs. Right now, it's not supported in Racetrack as it always pushes at least one image.

Plus, it might be surprising for the users that they can only deploy samples in this scenario. Out of curiosity, they would try to deploy something else, and quickly they discover that such Racetrack is useless.

kustomize/aks/dashboard.yaml
LookACastle: I think it's fine from a documentation standpoint then. Double checking, we do in fact recommend changing these for deploying to production, right? Because otherwise they wouldn't really be any good for authenticating anyone.

docs/deployment/k8s-installation.md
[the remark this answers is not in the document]
LookACastle: That's fair.

kustomize/aks/dashboard.yaml
JosefAssadERST: I'm not crazy about having a token here either, but I can see how automating it away might be a whole heap of new complexity.

What a project like Django does with its `SECRET_KEY` is, it autogenerates it.

Maybe a short-term low-tech option here is to leave it blank and instruct the user to fill it out during deployment?

iszulcdeepsense: Okay, I'm gonna add a section about how to generate these keys and where to fill it out. 

Django generates `SECRET_KEY` when creating a new project. We don't have that step and we have to mention that step in the docs.

iszulcdeepsense: I included steps for generating secrets and tokens.

kustomize/aks/image-builder.yaml
```diff
@@ -0,0 +1,92 @@
...
+            - name: CONFIG_FILE
+              value: /etc/racetrack/image-builder/config.yaml
+            - name: DOCKER_TLS_CERTDIR
+              value: ''
+            - name: DOCKER_HOST
+              value: tcp://127.0.0.1:2375
+            - name: DEBUG
+              value: 'true'
+            - name: LIFECYCLE_AUTH_TOKEN
+              value: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzZWVkIjoiZGVhNzYzMDAtN2JhYy00ODRjLTkyOTYtNWQxMGNkOTM3YTU4Iiwic3ViamVjdCI6ImltYWdlLWJ1aWxkZXIiLCJzdWJqZWN0X3R5cGUiOiJpbnRlcm5hbCIsInNjb3BlcyI6WyJmdWxsX2FjY2VzcyJdfQ.ND3wDeK58L5T1jIYcuArQ5O3M0Ez3_pCAEi5NXD_hLY'
```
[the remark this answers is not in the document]
iszulcdeepsense: I included steps for generating secrets and tokens. This has been transformed into a placeholder.

kustomize/aks/pub.yaml
iszulcdeepsense: I included steps for generating secrets and tokens. This has been transformed into a placeholder.

iszulcdeepsense:
@LookACastle  @JosefAssadERST  Does it look good to you now?

LookACastle (approved):
Yup, good work.

PR #223, opened 2023-03-24
Documentation update for job type development

### LINKED ISSUE #199: Documentation for job plugin development is too sparse
The job plugin development docs are too sparse, and expect a prospective developer to know a lot of stuff ahead of time - I think we should point them in the right direction a bit more.

"_You MAY implement swagger documentation for your endpoints on root endpoint._" should really have documentation for how to make a swagger endpoint. A link to the documentation on their side is probably adequate.

"_You MAY implement /metrics endpoint for exposing Prometheus metrics._" should also have documentation, exactly like described above.

It'd also be nice if, like the required steps, these optional steps had example code written for them. 

The "Overall workflow" is very narrow to only the initial concept of racetrack. It needs to be expanded to be understandable to a wider audience, if we want to keep it around at all. Not every racetrack user has to have data scientists. Kubernetes is not the only infrastructure supported by racetrack. So on, so forth.

LookACastle:
Switched from rust job type examples to go, adding swaggerui and prometheus code examples. 

Also ripped out "overall flow" as it assumes IKP's workflow as universal. It could also be reworked - but I doubt it even fits in this document to begin with. I've kept image file around in the repo in case someone decides it fits somewhere else.

Closes #199

### DIFF
--- a/docs/development/plugins-job-types.md
+++ b/docs/development/plugins-job-types.md
@@ -3,43 +3,25 @@
 This tutorial shows  how to create a plugin extending Racetrack with your own job types.
 Job types allow you to run applications written in specific programming languages.
 
-## Overall flow
-
-Here's the overall flow which describes how job types are created
-and used by others to deploy Job workloads:
-
-[![Job Type Plugins](../assets/job-type-plugins.png)](../assets/job-type-plugins.png)
-
-1. **Plugin Developer** writes the source code of a wrapper and base.Dockerfile.
-  He creates plugin code, Job template Dockerfile and plugin Manifest.
-  Next, he turns it into a ZIP plugin using a plugin bundler.
-2. He uploads the plugin to the Racetrack.
-3. **Data Scientist** writes the source code of his Job and pushes it to a Git repository.
-4. He creates job.yaml Manifest (including job type name and reference to a git repository)
-  and starts a deployment using racetrack client.
-5. Racetrack builds the base image and the Job image made of templated Job Dockerfile
-  and the Job type source code pulled from git.
-6. Racetrack creates Kubernetes resources for the Job workload.
-
 ## How to create a job-type plugin
 
-As an example, let's make a plugin to run jobs written in [Rust programming language](https://www.rust-lang.org/).
+As an example, let's make a plugin to run jobs written in [Go programming language](https://www.go.dev/).
 
 ### 1. Create a git repository
 Create a git repository (or use existing one) to keep the source code of the plugin.
 
-We're going to use https://github.com/TheRacetrack/plugin-rust-job-type 
-repository and we'll place the plugin inside `rust-job-type` subdirectory.
+We're going to use https://github.com/TheRacetrack/plugin-go-job-type 
+repository and we'll place the plugin inside `golang-job-type` subdirectory.
 
 ### 2. Initialize plugin manifest
-Create `plugin-manifest.yaml` file in a `rust-job-type` subdirectory.
+Create `plugin-manifest.yaml` file in a `golang-job-type` subdirectory.
 It contains the metadata of the plugin, including name, version
 and the URL of the plugin home page.
 
 ```yaml
-name: rust-job-type
-version: 1.0.0
-url: https://github.com/TheRacetrack/plugin-rust-job-type
+name: golang-job-type
+version: 0.0.0
+url: https://github.com/TheRacetrack/plugin-go-job-type
 ```
 
 ### 3. Write the wrapper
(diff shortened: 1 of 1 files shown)


docs/development/plugins-job-types.md
```diff
@@ -1,269 +1,1013 @@
...
-version = "0.1.0"
-edition = "2021"
-
-[dependencies]
-actix-web = "4.0.1"
-serde = { version = "1.0.137", features = ["derive"] }
-log = "0.4.17"
-env_logger = "0.9.0"
-serde_json = "1.0.81"
+  <summary>File `go_wrapper/go.mod`</summary>
```
iszulcdeepsense, line 395: Maybe we could point to the link https://github.com/TheRacetrack/plugin-go-job-type/blob/master/golang-job-type/go_wrapper/go.sum instead of copying this huge file here, which makes it also harder to maintain. 

Same with the `metrics.go` and `openapi.go` files which are non-obligatory for making something a job type.
I'm afraid that too complex job type documentation might scare off the developers.

iszulcdeepsense (approved):

docs/development/plugins-job-types.md
```diff
@@ -1,482 +1,372 @@
...
-## Overall flow
+## Quickstart Step-by-step
+1. Create a git repository for your plugin
 
-Here's the overall flow which describes how job types are created
-and used by others to deploy Job workloads:
+2. Create a [plugin manifest](../manifest-schema.md) in a plugin subdirectory
 
-[![Job Type Plugins](../assets/job-type-plugins.png)](../assets/job-type-plugins.png)
+3. Write a wrapper for the software you are making a job type for
```
iszulcdeepsense, line 11: Nitpick: A dot improves reading here
```suggestion
3. Write a wrapper for the software you are making a job type for.
```

docs/development/plugins-job-types.md
```diff
@@ -1,482 +1,372 @@
...
+│   └── perform.go
+├── health.go
+├── main.go
+├── Makefile
+└── server.go
 ```
 
-Let's assume that the user's source code will be placed in `rust_wrapper/src/handler/` subfolder.
-It will be injected there by docker during building the image.
+`go_wrapper/src/handler/` is for handling the user's code, it will be
```
iszulcdeepsense, line 115:
```suggestion
`go_wrapper/handler/` is for handling the user's code, it will be
```

docs/development/plugins-job-types.md
```diff
@@ -1,482 +1,372 @@
...
+```go
 // This is just a stub for IDE.
-// It gets replaced by user's Job code in job-template.Dockerfile
-use std::collections::HashMap;
-use serde_json::{Value, json};
-
-pub fn perform(input: HashMap<String, Value>) -> Value {
-    json!(input)
-}
+// It gets replaced by user's Job code in wrappers/docker/golang/job-template.Dockerfile
```
iszulcdeepsense, line 120: This comment is a bit out-of-date
```suggestion
// It gets replaced by user's Job code in golang-job-type/job-template.Dockerfile
```

docs/development/plugins-job-types.md
```diff
@@ -1,482 +1,372 @@
...
-  directory: sample-rust-function
-```
-
-Don't forget to push the Job code to a git repository.
-
-Finally, we can deploy this Job to Racetrack:
-```bash
-racetrack deploy
-```
+And then run `racetrack plugin bundle` in the `go-job-type` directory.
```
iszulcdeepsense, line 372: How about adding a "Frequently Asked Questions" section at the bottom?
It may explain a difference between base image and job template, which in my opinion can be a source of a common confusion.


## Frequently Asked Questions
### What distinguishes a base image from a job template?

The job image creation process involves two stages:

1. **Preparing the base image** - 
  The base image comprises files that are common to all jobs, such as your wrapper code, and isn't tied to any specific job or manifest.

2. **Building a job from the template** - 
  Before a job is deployed, the job template is combined with a user-submitted manifest to create the final Dockerfile that will be used to build the job image.
  The job image is an extension of the base image, with the job source code added during the building process. This stage is unique to each job.

iszulcdeepsense (approved):
Now it looks surprisingly short and easy. That's incredibly good :+1:

docs/development/plugins-job-types.md
LookACastle: I just created an FAQ section, is it okay or would you rather I just took what you've written here?

iszulcdeepsense: It's alright.

PR #226, opened 2023-03-28
Allow setting cookie on non-HTTPS deployments

iszulcdeepsense:
Racetrack's cookie can work even on non-HTTPS deployments, eg. AKS

### DIFF
--- a/dashboard/dashboard/middleware.py
+++ b/dashboard/dashboard/middleware.py
@@ -53,6 +53,7 @@ class UserCookieMiddleWare(object):
 def set_auth_token_cookie(auth_token: str, response: HttpResponse):
     # if cookie value contains /, it would get unnecessarily double quoted, so encode
     value = quote(auth_token, safe="")
+    secure_cookie: bool = os.environ.get('ENFORCE_SECURE_COOKIE', 'false').lower() == 'true'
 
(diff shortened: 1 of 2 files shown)


dashboard/dashboard/middleware.py
```diff
@@ -70,7 +65,7 @@ def set_auth_token_cookie(auth_token: str, response: HttpResponse):
         max_age=60 * 60 * 24 * 31,  # one month of cookie lifetime (in seconds), also makes cookie persist between browser restarts
         samesite='lax',  # cookie is sent only when user access origin site or navigates to it, prevents CSRF
         httponly=True,  # prevents client-side JavaScript read access
-        secure=True)  # cookie is sent only when request is made with https (except on localhost)
+        secure=False)  # cookie is sent only when request is made with https (except on localhost)
```
Koshmaar, line 68: Comment is out-of-date, now that secure=Flse. But, maybe the feature would be env variable controlled? So that ie. if RACETRACK_ENFORCE_SECURE_COOKIE is false, then it will work as in this PR. But if it's true, then it will work as before.

iszulcdeepsense: Good point, added `ENFORCE_SECURE_COOKIE`

Koshmaar (approved):

PR #229, opened 2023-03-30
Fix linter warnings, reformat code

iszulcdeepsense:


racetrack_client/racetrack_client/plugin/bundler/bundle.py
```diff
@@ -62,9 +62,9 @@ def _load_plugin_manifest(manifest_file: Path) -> PluginManifest:
         return parse_yaml_datamodel(yaml_str, PluginManifest)
 
 
-def _write_plugin_manifest(zip, plugin_manifest: PluginManifest):
+def _write_plugin_manifest(_zip, plugin_manifest: PluginManifest):
```
Koshmaar, line 65: _zip is missing type.

iszulcdeepsense: Good catch, I'll fix it on the other branch as I've already merged it too bravely.

Issue #211, opened 2023-03-16
Jobs stay visible after deletion via the dashboard.


LookACastle:
If you delete a job from the dashboard,the job stays on the dashboard, leaving stale data visible until you refresh the page. Ideally there wouldn't be stale data - I.E, the job shouldn't be shown any longer once deleted.

JosefAssadERST:
Agree. Do we have the backend endpoint so the accordion can just be refreshed with an ajax call?

iszulcdeepsense:
There's no backend endpoint for just the accordion, but the whole page. It makes me think of React so that components get refreshed once something changed in a reactive way.

Issue #206, opened 2023-03-14
Add link to Grafana on the Dashboard


iszulcdeepsense:
These useful links could include Grafana as well:

![Screenshot from 2023-03-14 17-29-46](https://user-images.githubusercontent.com/45496492/225073059-22c4ec1b-0e17-4127-b940-a07ef96d5945.png)

JosefAssadERST:
Yes please!

If we have a per-job dashboard then it would be awesome to link to those from the panel you get when you click a job in the list of jobs accordion.

If we don't have a per-job dashboard maybe we should have one?

PR #242, opened 2023-05-10
Minor UI improvements for a Dashboard

iszulcdeepsense:
Closes https://github.com/TheRacetrack/racetrack/issues/211

Closes https://github.com/TheRacetrack/racetrack/issues/206

Minor UI improvements for a Dashboard:
- Annotate login form fields correctly so that browsers can prompt you to remember the password.
- Add horizontal scroll to jobs tree in case of a long job name
- Add inner loading to graph view
- narrower navbar
- Show status indicator (green/red/yellow) instead of full status badge on a jobs tree
- "Jobs" panel takes up the whole space available - no white bars on the sides
- job name is not repeated in the tree leafs
- Add sorting jobs by status - first faulty jobs
- Store current jobs ordering in local storage
- show number of jobs near family
- Refresh list of jobs after deleting one
- Link to job-related Grafana dashobard

### DIFF
--- a/dashboard/dashboard-front/src/components/jobs/JobStatus.vue
+++ b/dashboard/dashboard-front/src/components/jobs/JobStatus.vue
@@ -7,12 +7,13 @@ const props = defineProps({
 })
 const status: Ref<string> = computed(() => props.status || "")
 const color: Ref<string> = computed(() => {
-    if (status.value == "running") {
-        return "green"
-    } else if (status.value == "error") {
-        return "red"
-    } else {
-        return "orange"
+    switch (status.value) {
+        case "running": 
+            return "green"
+        case "error": 
+            return "red"
+        default: 
+            return "orange"
     }
 })
 </script>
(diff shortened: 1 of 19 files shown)


dashboard/dashboard-front/src/components/jobs/JobStatus.vue
```diff
@@ -1,12 +1,29 @@
...
+})
+const status: Ref<string> = computed(() => props.status || "")
+const color: Ref<string> = computed(() => {
+    if (status.value == "running") {
+        return "green"
+    } else if (status.value == "error") {
+        return "red"
+    } else {
+        return "orange"
+    }
```
mpolomdeepsense, line 16: Generally, it is preferred to use strict equality over the normal one.

```suggestion
    if (status.value === "running") {
        return "green"
    } else if (status.value === "error") {
        return "red"
    } else {
        return "orange"
    }
```

mpolomdeepsense: Or it could be changed to `switch case`

mpolomdeepsense (approved):

PR #244, opened 2023-05-15
Online editing of job manifest

iszulcdeepsense:


dashboard/dashboard-front/src/assets/main.css
```diff
@@ -14,3 +14,7 @@
 .x-overflow-any {
   overflow-wrap: anywhere;
 }
+
+.fill-parent {
+  min-height: inherit;
+}
```
mpolomdeepsense, line 20: No new line

dashboard/dashboard-front/src/utils/error.ts
```diff
@@ -7,6 +7,15 @@ export function extractErrorDetails(err: any): string {
...
             return data.type
+        if (err.message) {
+            const method = err.config?.method?.toUpperCase()
+            if (data) {
+                const dataJson = JSON.stringify(data)
+                return `${err.message}: ${method} ${err.config?.url}: ${dataJson}`
+            } else {
+                return `${err.message}: ${method} ${err.config?.url}`
+            }
+        }
```
mpolomdeepsense, line 18: Just a small thing; requires only one change in case the return message format changes in the future.

```suggestion
        if (err.message) {
            const method = err.config?.method?.toUpperCase();
            const message = `${err.message}: ${method} ${err.config?.url}`;
            if (data) {
                const dataJson = JSON.stringify(data);
                return `${message}: ${dataJson}`;
            };
            return message;
        }
```

PR #247, opened 2023-05-16
Show more details in case of racetrack-client HTTP error

iszulcdeepsense:
When HTTP error occurs due to Cloudflare's proxy, it's hard to debug when all we got is just:
```
expected JSON response, got "text/html; charset=UTF-8" for url
```
Let's print more details in such cases:
- status code
- method


LookACastle (approved):
LGTM

PR #248, opened 2023-05-17
Keep record of a caller by PUB

iszulcdeepsense:


pub/lifecycle_test.go
```diff
@@ -13,11 +13,11 @@ func TestAuthorizeCaller(t *testing.T) {
 
 	httpmock.RegisterResponder("GET", "http://localhost:7202/lifecycle/api/v1/auth/can-call-job/skynet/latest",
 		httpmock.NewStringResponder(200,
-			`{"id": "000", "name": "skynet", "version": "0.0.1", "status": "running", "create_time": 1000, "update_time": 1000, "manifest": null, "internal_name": "adder-v-0-0-1"}`))
+			`{"job": {"id": "000", "name": "skynet", "version": "0.0.1", "status": "running", "create_time": 1000, "update_time": 1000, "manifest": null, "internal_name": "adder-v-0-0-1"}, "caller": "bob"}`))
 
 	httpmock.RegisterResponder("GET", "http://localhost:7202/lifecycle/api/v1/auth/can-call-job/skynet/0.0.1/api/v1/perform",
 		httpmock.NewStringResponder(200,
-			`{"id": "000", "name": "skynet", "version": "0.0.1", "status": "running", "create_time": 1000, "update_time": 1000, "manifest": null, "internal_name": "adder-v-0-0-1"}`))
+			`{"job": {"id": "000", "name": "skynet", "version": "0.0.1", "status": "running", "create_time": 1000, "update_time": 1000, "manifest": null, "internal_name": "adder-v-0-0-1"}, "caller": "bob"}`))
```
Koshmaar, line 20: Do you have here test for empty caller, ie for un-authorized calls? Would be useful to see that nothing breaks also in this case.

sample/python-logger/entrypoint.py
```diff
@@ -0,0 +1,19 @@
...
+        caller_name = self.get_caller_name()
+        logger.info(f'perform called by {caller_name}')
+        return random.random()
+
+    def get_caller_name(self) -> Optional[str]:
+        if hasattr(self, 'request_context'):
+            request = getattr(self, 'request_context').get()
+            return request.headers.get('X-Caller-Name')
+        else:
+            return None
```
Koshmaar, line 19: That's a bit unfriendly... maybe you could inject the caller name directly to envs that are available in perform? So user can just do `os.environ.get("CALLER_NAME")`. If not or that's a lot of work, ignore this. Users should be content with names in logs, AFAIK there's no use case for getting name in perform, and if they need it, they can copy-paste this, just as with calling other jobs.

Koshmaar (approved):
LGTM.

sample/python-logger/entrypoint.py
iszulcdeepsense: I'm afraid it might not work correctly, since `os.environ` is global and if we have multiple concurrent requests, it can lead to printing spurious data. `request_context` is a bit different, it's based on https://docs.python.org/3/library/contextvars.html, which guarantees to have a correct value, event in multi-threaded context.

Koshmaar: Good point. Didn't see this request_context before. One more question, why can't we depend on it being in job? Isn't it RT python server which is setting it, so theoretically it should always be there?

iszulcdeepsense: Not sure if I understand. The caller name changes over time. One time the caller is "alice", another time it's "bob". We're not sure what we get, if both requests happens at the same time. We risk with a race condition to happen, if we're gonna keep that information in a single variable.

Koshmaar: I mean why the code can't look like that:
```
    def get_caller_name(self) -> Optional[str]:
            request = getattr(self, 'request_context').get()
            return request.headers.get('X-Caller-Name')
```           
Not that I'm recommending such a change if there's even small chance to request_context be not accessible - have no idea where it's from, that's why I'm asking. Also, it's a nitpick, let it not block this PR.

iszulcdeepsense: Ah, I see now. You're right, it can be shortened.

Issue #231, opened 2023-04-11
Job manifests should be validated and universal


LookACastle:
Job manifests are currently not validated properly, leading to both potential messes in the future, and some current minor annoyances with racetrack. Their rules are also not universal for all job types (for example, python has special rules, as does golang, de facto limiting job type development).

For example, if someone produces a job type for a language where they'd like some customization of some sort, they have to either homebrew a solution outside of the [spec](https://github.com/TheRacetrack/racetrack/blob/b2ff0e29994e907f121ee00de8d287783dbed8d0/docs/manifest-schema.md) (For example, see [the Hugo jobtypes manifests](https://github.com/TheRacetrack/plugin-hugo-job-type/blob/0e97b2d3d09ca4c1fc5723d3d59431fa194e0472/sample/job.yaml) which has `wrapper_properties`), or they have to touch the main racetrack repository like we've done for `go` with the `golang` key, and `python` with the `python` key.

You can also have "wrong" manifests that follow our current schema. For example, if you make a job manifest with `name: no`, you get an error that the repository name must be lowercase. If you know YAML, you know that `no` becomes `False`, but it's not intuitive, and better validation would allow a better error message to occur, "`name` is of type boolean, it must be of type string" or similar. 

---

I propose that we make the spec universal, such that there's no more special job type keys, `lang` better reflect that job types don't have to be languages, and add a universal `jobtype-specific` key or similar, which is a YAML list that holds whatever the job type developer wants it to hold. `jobtype-specific` will be the replacement for fields like `python` and `golang`

I further propose that we introduce 2 steps of validation, one for the universal YAML, and one for the job type specific list. Racetrack core can validate everything outside of the jobtype specific key, and the job type developer should make sure their own specific fields are validated. 

The validation could be done using some schema tool like [Yamale](https://github.com/23andMe/Yamale), though I personally have no strong opinion on exactly how we do it, as long as it gets done.

JosefAssadERST:
Seeing how it's me that bullied @LookACastle into posting this, here's how I interpret the proposal in a mockup:

# Python example

```yaml
---
# CURRENT
name: meow
owner_email: meow@example.com
lang: python3:latest
git:
  remote: https://example.com/yeehaw
  branch: master
 python:
  requirements_path: 'supersmart/requirements.txt'
  entrypoint_path: 'job_entrypoint.py'
```

```yaml
---
# PROPOSED
name: meow  # validated by RT core
[...]

PR #249, opened 2023-05-17
`jobtype_extra` added to manifests, and schema validation has been added.

LookACastle:
This pull request will unify `wrapper_properties`, `golang`, and `python` fields into one, so as to support future developers. It also adds schema validation as opposed to the `pydantic` way we used before.

Known issues:

Ideally we'd create some mechanism for jobtypes to simply have to supply a schema to get validation as well, but that seems non-trivial at this point, and the merge request is already quite large. For now, it's the job of the jobtypes themselves to implement if they want manifest validation. I'll likely open a follow-up issue unless the reviewer has any insights making it trivial. 

I do not think this closes the [original issue](https://github.com/TheRacetrack/racetrack/issues/231), and have posted a comment there to explain why.


docs/job_types/job_golang.md
```diff
@@ -2,7 +2,7 @@
 "golang" job type is intended to handle the calls to your Go function.
 Racetrack will wrap it up in a web server.
 
-Set `lang: golang:latest` in your `job.yaml` manifest file in order to use this type of job.
+Set `jobtype: golang:latest` in your `job.yaml` manifest file in order to use this type of job.
```
iszulcdeepsense, line 5: Let's get rid of this file. It's been moved to https://github.com/TheRacetrack/plugin-go-job-type/blob/master/docs/job_golang.md, but for some reason this leftover remained.

docs/job_types/job_python3.md
```diff
@@ -2,7 +2,7 @@
 "python3" job type is intended to handle the calls to your Python function.
 Racetrack will wrap it up in a web server.
 
-Set `lang: python3:latest` in your `job.yaml` manifest file in order to use this type of job.
+Set `jobtype: python3:latest` in your `job.yaml` manifest file in order to use this type of job.
```
iszulcdeepsense, line 5: Let's remove job_python3.md in here. It's been moved to https://github.com/TheRacetrack/plugin-python-job-type/blob/master/docs/job_python3.md

racetrack_client/racetrack_client/manifest/validate.py
```diff
@@ -30,17 +32,11 @@ def load_validated_manifest(path: str) -> Manifest:
...
-
-    assert 1 <= manifest.replicas <= 15, 'replicas count out of allowed range'
-
-    assert urlsplit(manifest.git.remote).scheme == 'https', 'git remote URL should be HTTPS'
-
+    serialized_manifest = manifest.dict(exclude_none=True)
+    path = Path(__file__).with_name('schema.json')
+    with path.open('r') as f:
+        schema = json.load(f)
+        validate(serialized_manifest, schema=schema)
```
iszulcdeepsense, line 39: I think we can leave job version validation. Other places in code expect version to be semantic and it may fail if we're going to accept any string.
```suggestion
        validate(serialized_manifest, schema=schema)
        
    with wrap_context('parsing Job version'):
        SemanticVersion(manifest.version)
```

racetrack_client/racetrack_client/manifest/manifest.py
```diff
@@ -108,8 +90,19 @@ class Manifest(BaseModel, extra=Extra.forbid, arbitrary_types_allowed=True):
...
+    jobtype_extra: Optional[Dict[str, Any]] = None
+    golang: Optional[Dict[str, Any]] = None #Deprecated
+    python: Optional[Dict[str, Any]] = None #Deprecated
+    wrapper_properties: Optional[Dict[str, Any]] = None #Deprecated
 
     # Back-end platform where to deploy the service
     infrastructure_target: Optional[str] = None
+
+    def get_jobtype(self):
+        return self.jobtype if self.lang is None else self.lang
```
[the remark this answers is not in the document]
LookACastle: I agree, but will note on the side that they cannot fill up both fields with the schema validation.

racetrack_client/racetrack_client/manifest/validate.py
LookACastle: I also added this check to the schema.

iszulcdeepsense (approved):

PR #251, opened 2023-05-22
Keep manifest up-to-date with running job

iszulcdeepsense:
This runs full redeployment right after editing the manifest of a job.

### DIFF
--- a/dashboard/dashboard-front/src/components/jobs/JobDetails.vue
+++ b/dashboard/dashboard-front/src/components/jobs/JobDetails.vue
@@ -27,8 +27,10 @@ const loadingLogs: Ref<boolean> = ref(false)
 const manifestDialogRef: Ref<typeof ManifestEditDialog | null> = ref(null)
 
 const manifestYaml: Ref<string> = computed(() => {
-    if (!job.value?.manifest)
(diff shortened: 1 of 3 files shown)


dashboard/dashboard-front/src/components/jobs/JobDetails.vue
```diff
@@ -26,9 +26,11 @@ const logsOpen: Ref<boolean> = ref(false)
 const loadingLogs: Ref<boolean> = ref(false)
 const manifestDialogRef: Ref<typeof ManifestEditDialog | null> = ref(null)
 
-const manifestYaml: Ref<string> = computed(() => 
-    yaml.dump(removeNulls(job.value?.manifest)) || ''
-)
+const manifestYaml: Ref<string> = computed(() => {
+    if (!job.value?.manifest)
```
LookACastle, line 30: This is if a job does not have a manifest, correct? This doesn't seem like something that'd happen under normal use, right? If so, it's maybe worth logging a warning.

LookACastle (commented):
A single minor question, other than that, looks good.

LookACastle (approved):
LGTM

PR #254, opened 2023-05-24
Use latest python plugin in E2E tests

iszulcdeepsense:


LookACastle (approved):
This seems like my mess your cleaning up, sorry about that :sweat_smile:. LGTM though, nice that we're putting it into constants so it's harder to make the same mistake in the future, rather than just bumping the version.

PR #263, opened 2023-06-02
Keep original manifest YAML

iszulcdeepsense:


lifecycle/lifecycle/deployer/provision.py
```diff
@@ -64,10 +64,12 @@ def provision_job(
 
         job: JobDto = job_deployer.deploy_job(manifest, config, plugin_engine,
                                               tag, runtime_env_vars, family_dto, containers_num)
-        job.deployed_by = deployment.deployed_by
-        job.job_type_version = f'{job_type.lang_name}:{job_type.version}'
 
     with wrap_context('saving job in database'):
+        job.deployed_by = deployment.deployed_by
+        job.manifest = manifest
```
LookACastle, line 70: Is job not already loaded with this manifest at line 65?

LookACastle (commented):
A general comment to a lot of places in the code:

It feels odd having manifest_yaml being passed around as a parameter next to manifest. Doubly so when you've created convert_to_yaml. Would it maybe make more sense to just convert the manifest where needed, rather than passing the extra parameter around? It'll probably look a lot cleaner.

LookACastle:
Would it be doable to have the `Manifest` object keep a field like `Manifest.originator_yaml` or something similar, which would hold the data `manifest_yaml` currently holds?

iszulcdeepsense:
> Would it be doable to have the `Manifest` object keep a field like `Manifest.originator_yaml` or something similar, which would hold the data `manifest_yaml` currently holds?

Good idea (if it's doable). I think pydantic allows for having hidden fields that are not exported so it could be the good place to go https://docs.pydantic.dev/latest/usage/exporting_models/#model-and-field-level-include-and-exclude

iszulcdeepsense:
> > Would it be doable to have the `Manifest` object keep a field like `Manifest.originator_yaml` or something similar, which would hold the data `manifest_yaml` currently holds?
> 
> Good idea (if it's doable). I think pydantic allows for having hidden fields that are not exported so it could be the good place to go https://docs.pydantic.dev/latest/usage/exporting_models/#model-and-field-level-include-and-exclude. I'll check on that

@LookACastle  Alright, I moved `manifest_yaml` to `Manifest._origin_yaml` field, thus getting rid of a doubled parameter. However, `JobDto` class remained with twofold manifests as it is a data structure being sent to front-end and it needs both formats there.

LookACastle (approved):

PR #267, opened 2023-06-13
Use QTable in Portfolio tab

iszulcdeepsense:
Dashboard now renders a table on the portfolio tab with QTable rather than TableFilter,
  which occasionally seemed to have rendering issues.
  Plus, it's more consistent with the current UI styling.

This is how Portfolio looks now with QTable from Quasar components:
Screenshot from 2023-06-13 14-18-17


mpolomdeepsense (approved):
Looking good.

One interesting thing that I noticed is between current lines 48-50 in PortfolioTable.vue where you make the API call. You use `as PortolioJob[]` but methods of `apiClient` are generic and you should be able to make something like this `apiClient.get<PortolioJob[]>(...)`. Not sure if it's fully correct since there is AxiosResponse<any> as a default type so maybe `apiClient.get<AxiosResponse<PortolioJob[]>>(...)` or even better to hide `AxiosResponse` somewhere inside the methods.

Nothing bad with using `as` there, but it always felt to me like generics in situations like that is more pleasing to the eye.

PR #268, opened 2023-06-13
Redirect to login page when session expires

iszulcdeepsense:


dashboard/dashboard-front/src/services/UserDataStore.ts
```diff
@@ -81,3 +83,17 @@ function saveUserData() {
...
+
+function getCookie(name: string): string | null {
+    const nameLenPlus = name.length + 1
+    const values = document.cookie
+        .split(';')
+        .map(c => c.trim())
+        .filter(cookie => cookie.substring(0, nameLenPlus) === `${name}=`)
+        .map(cookie => decodeURIComponent(cookie.substring(nameLenPlus)))
+    return values[0] || null
+}
```
mpolomdeepsense, line 99: I was confused at the beginning why there was a `+ 1`. I feel like this change makes it a little easier to understand the code.

```suggestion
function getCookie(name: string): string | null {
    const nameWithSuffix = `${name}=`;
    const values = document.cookie
        .split(';')
        .map(c => c.trim())
        .filter(cookie => cookie.substring(0, nameWithSuffix.length) === nameWithSuffix)
        .map(cookie => decodeURIComponent(cookie.substring(nameWithSuffix.length)))
    return values[0] || null
}
```

dashboard/dashboard-front/src/services/UserDataStore.ts
```diff
@@ -81,3 +83,17 @@ function saveUserData() {
         localStorage.setItem('userData.authToken', authToken.value)
     }
 }
+
+export function hasAuthCookie(): boolean {
+    return getCookie(AUTH_COOKIE_NAME) != null
```
mpolomdeepsense, line 88: It's impossible for `getCookie` function to return anything else than a `null` or a `string` and the string can't be empty because of `... || null` so I would go with strict equality here.

```suggestion
    return getCookie(AUTH_COOKIE_NAME) !== null
```

PR #289, opened 2023-07-25
Add notice about Chain calls in jobtype docs

iszulcdeepsense:


docs/development/plugins-job-types.md
```diff
@@ -68,6 +68,9 @@ Every wrapper has to follow some rules:
 - You MAY implement [swagger](https://swagger.io/tools/swagger-ui/) documentation for your endpoints on root endpoint.
 - You MAY implement `/metrics` endpoint for exposing [Prometheus](https://prometheus.io/) metrics.
 - You MAY expose any other endpoints.
+- Chain calls to other jobs SHOULD be made by
```
LookACastle, line 71: Is it obvious to an outside reader what a "chain call" is? I find it quite intuitive, but maybe something like this is more understandable to those not in the know
```suggestion
- Calls from jobs to other jobs SHOULD be made by
```

LookACastle (approved):
A single comment, approving anyhow because it's simply a thing to ponder, rather than anything I perceive as an error.

PR #290, opened 2023-07-26
Replace old version when uploading a plugin

iszulcdeepsense:
- Racetrack-client has new flag `--replace` available when uploading a plugin:
  `racetrack plugin install <file.zip> --replace`.
  It will delete the existing versions (with the same name) on uploading a new one.
  Accordingly, there is a checkbox on the Dashboard's *Administration* tab for replacing a plugin on upload.
  By default, uploading a new plugin doesn't replace the older versions.

Screenshot from 2023-07-27 09-17-46

### DIFF
--- a/dashboard/dashboard-front/package-lock.json
+++ b/dashboard/dashboard-front/package-lock.json
@@ -2062,9 +2062,9 @@
       }
     },
     "node_modules/semver": {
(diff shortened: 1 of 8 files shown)


LookACastle (approved):
LGTM

PR #292, opened 2023-07-26
Load plugins selectively by declared components only

iszulcdeepsense:
- Skip loading a plugin if its intended to be loaded only by particular Racetrack components (eg. only by Lifecycle, not image-builder), if it's declared in plugin's manifest.
  It avoids doing this ugly trick https://github.com/TheRacetrack/plugin-kubernetes-infrastructure/blob/2dd9e2c2a5085cc7b921ab06ccf3603a2d74035c/src/plugin.py#L7C1-L7C1
- add plugin's labels - textual keywords describing kind of the plugin, e.g. 'infrastructure', 'job-type'

Example plugin-manifest.yaml:
```yaml
name: docker-infrastructure
version: 1.3.1
url: https://github.com/TheRacetrack/plugin-docker-infrastructure
components:
  - 'lifecycle'
category: 'infrastructure'
```

Screenshot from 2023-07-27 09-24-39

### DIFF
--- a/docs/development/developing-plugins.md
+++ b/docs/development/developing-plugins.md
@@ -40,6 +40,8 @@ It can have the following fields in YAML format:
 - `version` (**required**) - version of the plugin
 - `url` (optional) - a link to the plugin page
 - `priority` (optional) - order in plugins sequence, lowest priority gets executed first. Integer field, 0 by default.
+- `category` (optional) - kind of the plugin, either 'infrastructure', 'job-type' or 'core'
+- `components` (optional) - list of Racetrack components that the plugin should be running on, e.g. 'lifecycle', 'image-builder'
 
 Example:
 ```yaml
@@ -47,6 +49,9 @@ name: skynet-watcher
 version: 1.2.3
 url: https://github.com/TheRacetrack/racetrack
 priority: -1
+category: core
+components:
+  - lifecycle
 ```
 
 ### Building a plugin
--- a/docs/development/developing-plugins.md
+++ b/docs/development/developing-plugins.md
@@ -117,3 +117,9 @@ def post_job_delete(self, job: JobDto, username_executor: str = None):
     :param username_executor: username of the user who deleted the job
     """
 ```
+
+- `run_action` - Call a supplementary action of a plugin
+```python
+def run_action(self, **kwargs) -> Any:
+    """Call a supplementary action of a plugin"""
+```
--- a/docs/development/developing-plugins.md
+++ b/docs/development/developing-plugins.md
@@ -41,7 +41,8 @@ It can have the following fields in YAML format:
 - `url` (optional) - a link to the plugin page
 - `priority` (optional) - order in plugins sequence, lowest priority gets executed first. Integer field, 0 by default.
 - `category` (optional) - kind of the plugin, either 'infrastructure', 'job-type' or 'core'
-- `components` (optional) - list of Racetrack components that the plugin should be running on, e.g. 'lifecycle', 'image-builder'
+- `components` (optional) - list of Racetrack components that the plugin should be running on, e.g. 'lifecycle', 'image-builder'.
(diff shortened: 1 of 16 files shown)


iszulcdeepsense:
Instead of multi-value labels, let's use single enum type:
```yaml
category: infrastructure|job-type|core
```

racetrack_commons/racetrack_commons/plugin/core.py
```diff
@@ -56,6 +56,10 @@ def markdown_docs(self) -> str | None:
         """
         return None
 
+    def run_action(self, **kwargs) -> Any:
+        """Call a supplementary action of a plugin"""
+        return None
```
Koshmaar, line 61: This is interesting - what kind of supplementary action would be example here? Also, it's not mentioned in the changelog and documentation.

lifecycle/lifecycle/endpoints/plugin.py
```diff
@@ -70,6 +70,12 @@ def _info_plugin_docs(plugin_name: str) -> Optional[str]:
         """Get documentation for this plugin in Markdown format"""
         return plugin_engine.invoke_one_plugin_hook(plugin_name, PluginCore.markdown_docs)
 
+    @api.get('/plugin/{plugin_name}/run')
+    def _run_plugin_action(plugin_name: str, request: Request) -> Any:
+        """Call a supplementary action of a plugin"""
+        params = dict(request.query_params)
+        return plugin_engine.invoke_one_plugin_hook(plugin_name, PluginCore.run_action, **params)
```
Koshmaar, line 77: Hmm, `run` action sounds like a POST method.

docs/development/developing-plugins.md
```diff
@@ -40,13 +40,18 @@ It can have the following fields in YAML format:
 - `version` (**required**) - version of the plugin
 - `url` (optional) - a link to the plugin page
 - `priority` (optional) - order in plugins sequence, lowest priority gets executed first. Integer field, 0 by default.
+- `category` (optional) - kind of the plugin, either 'infrastructure', 'job-type' or 'core'
+- `components` (optional) - list of Racetrack components that the plugin should be running on, e.g. 'lifecycle', 'image-builder'
```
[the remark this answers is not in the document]
iszulcdeepsense: Right, If it's empty, plugins are loaded on all components. I'll make a note.

docs/development/developing-plugins.md
```diff
@@ -40,13 +40,18 @@ It can have the following fields in YAML format:
 - `version` (**required**) - version of the plugin
 - `url` (optional) - a link to the plugin page
 - `priority` (optional) - order in plugins sequence, lowest priority gets executed first. Integer field, 0 by default.
+- `category` (optional) - kind of the plugin, either 'infrastructure', 'job-type' or 'core'
```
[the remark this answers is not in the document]
iszulcdeepsense: For now it is true - it's just a label. Nevertheless, there are plans to treat plugins differently when uploading them (replace or leave the old version depending on the category). So I wouldn't guarantee that.
In this merge request, I'm just introducing this field. Then, we'll have to adapt plugins to use it. And afterwards, we can change the treatment, as a separate change.

lifecycle/lifecycle/endpoints/plugin.py
iszulcdeepsense: Yes, definitely.

docs/development/developing-plugins.md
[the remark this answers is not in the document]
Koshmaar: Fair enough.

racetrack_commons/racetrack_commons/plugin/core.py
Koshmaar: Cool :)

Koshmaar (approved):

PR #296, opened 2023-08-09
Finalize database probe queries

iszulcdeepsense:


lifecycle/lifecycle/endpoints/health.py
```diff
@@ -58,7 +58,9 @@ def is_database_connected() -> bool:
             shell(f'pg_isready -h {host} -p {port} -U {user} -d {db_name}', print_stdout=False)
 
         close_old_connections()
-        connection.cursor().execute("select 1")
+        with connection.cursor() as cursor:
+            cursor.execute('select 1')
+            cursor.fetchone()
```
Koshmaar, line 63: Maybe sth like `connection.close()`? I know it's possible in psycopg2 direclty, but don't know if it will work here with this django wrapper.

iszulcdeepsense:
@Koshmaar  Here's a fun thing:
I tried this:
```python
        with connection.cursor() as cursor:
            cursor.execute('select 1')
            cursor.fetchone()
            cursor.execute('select 2')
            cursor.fetchone()
```
and I ended up with
`513	2.792784	2023-08-10 13:43:42.164 +0200	racetrack	select 2	idle`
and no `select 1` in `pg_stat_activity`,
so it looks like the session is truly reused by Django, and `pg_stat_activity` shows the last query within an active session (not necessarily being processed right now)

Koshmaar (approved):
LGTM

PR #297, opened 2023-08-10
Add demo video

iszulcdeepsense:
Check this out, here's Marek's demo: https://github.com/TheRacetrack/racetrack/blob/286-video-tutorials-on-how-to-setup-use-add-jobs-etc/README.md
I put the short video in the main Readme page, also a longer video with subtitles is added here: https://github.com/TheRacetrack/racetrack/blob/286-video-tutorials-on-how-to-setup-use-add-jobs-etc/docs/user.md

Plus, I added a script for recreating the video in future


LookACastle (approved):
LGTM

PR #323, opened 2023-09-28
Fix deleting failing plugins

iszulcdeepsense:
It handles the case when you want to delete a faulty plugin that is not on the list of the loaded plugins (because it failed to load), but the files still remain.

### DIFF
--- a/racetrack_client/racetrack_client/utils/semver.py
+++ b/racetrack_client/racetrack_client/utils/semver.py
@@ -70,6 +70,11 @@ class SemanticVersion:
         latest_version = max(versions_objects.keys())
         return versions_objects[latest_version]
 
+    @property
+    def is_stable(self) -> bool:
+        """Return true, if it's not a pre-release version - has no pre-release label."""
+        return not bool(self.label)
+
     def __str__(self):
         return f'{self.major}.{self.minor}.{self.patch}{self.label}'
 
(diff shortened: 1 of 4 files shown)


racetrack_client/racetrack_client/utils/semver.py
```diff
@@ -70,6 +70,11 @@ def find_latest_wildcard(
         latest_version = max(versions_objects.keys())
         return versions_objects[latest_version]
 
+    @property
+    def is_stable(self) -> bool:
+        """Return true, if it's not a pre-release version - has no pre-release label."""
+        return not bool(self.label)
```
mpolomdeepsense, line 76: This also works

```suggestion
        return not self.label
```

image

racetrack_commons/racetrack_commons/plugin/engine.py
```diff
@@ -226,37 +230,50 @@ def write_plugin_config(self, name: str, version: str, config: str):
...
-        else:
-            logger.warning(f'ZIP plugin was not found: {plugin_data.zip_path}')
+            if self._delete_plugin_files(plugin_data.zip_path.stem):
+                logger.info(f'Older plugin version has been deleted: {plugin_data.zip_path.name}')
+            else:
+                logger.warning(f'Plugin files were not found: {plugin_data.zip_path.stem}')
 
-        extracted_dir = Path(self.plugins_dir) / EXTRACTED_PLUGINS_DIR / plugin_data.zip_path.stem
+    def _delete_plugin_files(self, stem: str) -> bool:
+        extracted_dir = Path(self.plugins_dir) / EXTRACTED_PLUGINS_DIR / stem
```
mpolomdeepsense, line 267: I noticed that `plugins_dir` is casted to Path with each use, maybe it could be done once during object init.

mpolomdeepsense (approved):

PR #329, opened 2023-10-02
Improve checking database connection status

iszulcdeepsense:
- switched to psycopg in version 3
- It checks database connection in background to avoid blocking metrics endpoint.
Scraping metrics no longer calls database query, but it reads from a cached result, which is updated periodically in a background thread.

In metrics endpoint stress test, it reduced number of open database connections from 23 to 11


lifecycle/lifecycle/server/db_status.py
```diff
@@ -0,0 +1,56 @@
...
+
+def is_database_connected() -> bool:
+    try:
+        django_db_type = os.environ.get('DJANGO_DB_TYPE', 'sqlite')
+        if django_db_type == 'postgres':
+            db_name = settings.DATABASES['default']['NAME']
+            user = settings.DATABASES['default']['USER']
+            host = settings.DATABASES['default']['HOST']
+            port = settings.DATABASES['default']['PORT']
+            shell(f'pg_isready -h {host} -p {port} -U {user} -d {db_name}', print_stdout=False)
```
Koshmaar, line 43: Why are you checking conn with pg_isready when you are also calling the connection check using cursor.execute? Isn't one enough?

Koshmaar (approved):
LGTM. The only thing missing is changelog, but you would do that probably as next step.

iszulcdeepsense:
> The only thing missing is changelog, but you would do that probably as next step.

I wasn't going to mention that in changelog as Racetrack's changelog contains only "**user-facing**, notable changes".

lifecycle/lifecycle/server/db_status.py
iszulcdeepsense: I'd strongly like to avoid making `select 1` queries, whenever I can, especially when the database is down and it leads to hanging the thread. That's why I probe the database with a different, lightweight check, which aborts much earlier. But it's not that definitive to tell there is a truly vital connection, and I need to make query anyways.

Koshmaar:
There can be many users of different sorts :) One is DS who are using RT, and they don't expect technical details. But there can be also people outside our organisation who use RT, there are backend engineers who pull newest commits and look into changelog looking for details, they are also our users.

lifecycle/lifecycle/server/db_status.py
Koshmaar: Ok, makes sense. However this made me think about the thread - if it fails, then the pod will soon fail all healthchecks, and pod will restart. Which is fine, then a new thread should start, so RT will still work (after few secs). You're catching all exceptions so it seems ok in this regard. Perhaps adding a timeout ie. 10s, would avoid it hang, and spare the pod restart.

iszulcdeepsense: > if it fails, then the pod will soon fail all healthchecks, and pod will restart

It won't. That's what this change was about. Losing connection to database will no longer restart the pod (it was never supposed to do that actually). This connection status is only printed as a metric and getting this metric is instant.

> Perhaps adding a timeout ie. 10s, would avoid it hang

I tried with `SET statement_timeout` to no avail. Moreover, there's no neat way in Python to timeout a hanging thread. Thread doesn't event have a method to kill it.

Koshmaar: I was actually thinking about other way the thread can fail, that is - just stop working completely due to some exception. So then the db connection status wouldn't be refreshed. But as described, it's rather impossible - every python exception will be caught by BaseException (except  KeyboardInterrupt and Python-raised errors like SystemExit, but it seems ok, first won't appear in pod, second probably means whole pod is being killed).

> It won't. That's what this change was about. Losing connection to database will no longer restart the pod (it was never supposed to do that actually). This connection status is only printed as a metric and getting this metric is instant.

Great. This however causes other thing when the thread hangs: it won't be able to report that there are connection problems to db. Health endpoint will return old "OK" data. 

> I tried with SET statement_timeout to no avail. Moreover, there's no neat way in Python to timeout a hanging thread. Thread doesn't event have a method to kill it.
But it's not a serious issue anymore. If it hangs, it hangs just this dedicated thread, so we'll just get information about broken connection a bit later.

Yep, the monitoring data will be untrue for some time (assuming that hanging will finish in minute or two, or even ten... no idea how long it lasts). Wouldn't https://github.com/litl/backoff max_time help here?

iszulcdeepsense: Ah, now I see what you meant.
 
> Wouldn't https://github.com/litl/backoff max_time help here?

Unfortunately, no. It would be helpful if we had to deal with something repetitive, giving back the control to the Python program. In this case, the whole interperer's thread hangs, not giving a chance for anything else to run, so adding more code won't help - it won't be executed anyways.

Koshmaar: Ah, whole interpreter hangs - interesting. Perhaps then we should rewrite Lifecycle to C++, so that we have more control over it. Ok, I'm joking ;) golang should be enough low-level, but that's IKP 2. Closing this thread.

PR #330, opened 2023-10-05
Provide instructions on how to install racetrack to a VM instance

iszulcdeepsense:


docs/deployment/standalone-host.md
```diff
@@ -0,0 +1,62 @@
...
+```sh
+sh <(curl -fsSL https://raw.githubusercontent.com/TheRacetrack/racetrack/master/utils/standalone-wizard/runner.sh)
+```
+Follow the installation steps. Shortly after, your Racetrack instance will be ready.
+
+### Notes
+
+- Pay attention to the output. The installer will generate unique passwords for your setup,
+  you'll be provided with the Dashboard address and superuser credentials.
+- Set the environment variable `export RT_NON_INTERACTIVE=1`
```
LookACastle, line 38: Is there a way to pass these questions as parameters so you can be non-interactive _and_ not go with the defaults? If yes, this should mention it, if no, we should maybe add that as an option.

iszulcdeepsense: Good point. In fact there is only one parameter right now. Added such option and included it in docs

LookACastle (approved):

PR #333, opened 2023-10-05
Add admin endpoint for cleaning up all plugins

iszulcdeepsense:


racetrack_commons/racetrack_commons/plugin/engine.py
```diff
@@ -244,6 +244,17 @@ def delete_plugin_by_version(self, name: str, version: str):
         self._record_last_change()
         self._load_plugins()
 
+    def delete_all_plugins(self):
+        for plugin_zip_path in sorted(self.plugins_path.glob('*.zip')):
+            plugin_zip_path.unlink()
+
+        extracted_dir = self.plugins_path / EXTRACTED_PLUGINS_DIR
+        if extracted_dir.is_dir():
+            shutil.rmtree(extracted_dir)
```
mpolomdeepsense, line 253: Not sure if it's possible here but it could potentially throw a permission error. There are `ignore_error` and `onerror` arguments for this method so maybe they could be used.

mpolomdeepsense (approved):

PR #334, opened 2023-10-06
Update grafana and PostgreSQL

iszulcdeepsense:


kustomize/external/pgbouncer.env
```diff
@@ -8,6 +8,7 @@ POOL_MODE=transaction
 MAX_CLIENT_CONN=30000
 CLIENT_IDLE_TIMEOUT=90
 ADMIN_USERS=postgres,admin
+AUTH_TYPE=scram-sha-256
```
Koshmaar, line 11: Why do you need this change? It seems the pgbouncer version on appDb hasn't changed. I understand it's to have same config in localhost and on production, and for production the change doesn't hurt.

Koshmaar (approved):

kustomize/external/pgbouncer.env
iszulcdeepsense: It turns out that latest PostgreSQL (16) uses different auth type by default, which is not enabled by default in latest pgbouncer (Yes, pgbouncer is incompatible with latest postgres). As long as we keep old postgres in AppDb (or with SCRAM passwords explicitly disabled), it won't hurt anything. Beware of upgrading prod's postgres to version 16 though.

Koshmaar: Good to know, thanks. Maybe write about it to Anders K, just to make sure he knows this.

PR #347, opened 2023-10-17
Fix intermittent error when uploading a plugin

iszulcdeepsense:
- Take available job types from lifecycle, not from image builder.
- Keep that in cache that call plugins engine only on-write


racetrack_commons/racetrack_commons/deploy/job_type.py
```diff
@@ -93,22 +95,24 @@ def gather_job_types(
...
 
     return job_types
 
 
-def list_available_job_types(plugin_engine: PluginEngine) -> list[str]:
-    with wrap_context('gathering available job types'):
-        job_types = gather_job_types(plugin_engine)
-    return sorted(job_types.keys())
+def _validate_job_type(job_type: JobType):
+    job_full_name = f'{job_type.lang_name}:{job_type.version}'
```
Koshmaar, line 114: It looks like a candidate for a property in JobType: `def job_full_name(): return ....`, because probably this kind of code is repeated in other places.

Koshmaar (approved):

PR #348, opened 2023-10-18
Override manifest values with command line params

iszulcdeepsense:
### Added
- Manifest values can be overriden with key-value pairs coming from a command line.
  Racetrack client has `--extra-vars KEY=VALUE` parameter (or `-e` in short)
  that overwrites values found in YAML manifest.

  - `KEY` is the name of field and it can contain dots to refer to a nested field, for example `git.branch=master`.
  - `VALUE` can be any YAML or JSON object.

  Extra vars parameters can be used multiple times in one command.  
  Example: `racetrack deploy -e secret_runtime_env_file=.env.local -e git.branch=$(git rev-parse --abbrev-ref HEAD)`  
  It makes CLI commands more script-friendly, so you can overwrite manifest without tracking changes in job.yaml file.  
  Tip: Use `racetrack validate` command beforehand to make sure your final manifest is what you expected.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Manifest values can be overriden with key-value pairs coming from a command line.
+  Racetrack client has `--extra-vars KEY=VALUE` parameter (or `-e` in short)
+  that overwrites values found in YAML manifest.
+  `KEY` is the name of field and it can contain dots to refer to a nested field, for example `git.branch=master`.
+  `VALUE` can be any YAML or JSON object. 
+  Extra vars parameter can be used multiple times in one command.
+  Example: `racetrack deploy -e secret_runtime_env_file=.env.local`.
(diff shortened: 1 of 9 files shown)


docs/CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  that overwrites values found in YAML manifest.
+
+  - `KEY` is the name of field and it can contain dots to refer to a nested field, for example `git.branch=master`.
+  - `VALUE` can be any YAML or JSON object.
+
+  Extra vars parameters can be used multiple times in one command.  
+  Example: `racetrack deploy -e secret_runtime_env_file=.env.local -e git.branch=$(git rev-parse --abbrev-ref HEAD)`  
+  It makes CLI commands more script-friendly, so you can overwrite manifest without tracking changes in job.yaml file.  
+  Tip: Use `racetrack validate` command beforehand to make sure your final manifest is what you expected.
+  ([#340](https://github.com/TheRacetrack/racetrack/issues/340))
```
mpolomdeepsense, line 20: Maybe it would be nice to state that it doesn't override the manifest file, just in-memory. From what I understood if you reprovision the job it will reuse the in-file config and not the one in memory, so it would be nice to add that.

mpolomdeepsense (approved):

PR #356, opened 2023-10-27
Add installer steps for remote kubernetes/docker infra

iszulcdeepsense:


utils/standalone-wizard/remote-kubernetes/remote-pub.yaml
```diff
@@ -0,0 +1,79 @@
...
+    metadata:
+      labels:
+        app.kubernetes.io/name: pub
+    spec:
+      serviceAccountName: null
+      securityContext:
+        supplementalGroups: [200000]
+        fsGroup: 200000
+        runAsUser: 100000
+        runAsGroup: 100000
```
Koshmaar, line 23: I wonder if those are needed also when running RT outside Kubernilla context... probably not, but there is a user 100k hardcoded in the dockerfile. I'm not sure if it's bad, ie. perhaps someone could want to run RT in their own hardened K8s env, which enforces other things (like user ids) than our prod env. But I can be completely wrong and perhaps this is neutral (user is created in dockerfile and K8s just uses it). It could be helpful to consult some devops, with more experience in different envs.

utils/standalone-wizard/wizard.py
Koshmaar, line 1: It's quite long - how about separating it to few files, ie. by target type..

utils/standalone-wizard/wizard.py
```diff
@@ -171,17 +452,33 @@ def _generate_secrets(config: 'SetupConfig'):
...
-def get_admin_auth_token(password: str) -> str:
-    response = Requests.post('http://127.0.0.1:7103/dashboard/api/v1/users/login', json={
+def login_first_time(config: 'SetupConfig', lifecycle_url: str):
+    try:
+        auth_token = get_admin_auth_token('admin', lifecycle_url)
+        logger.info('Changing default admin password…')
+        change_admin_password(auth_token, 'admin', config.admin_password, lifecycle_url)
+    except ResponseError as e:
+        if not e.status_code == 401:  # Unauthorized
+            raise e
```
Koshmaar, line 462: Nitpick, it could be written more succintly as `if e.status_code != 401`. But I wonder what is the intention here, because if it's 401, then the response error won't be re-raised. So the further functions will be called, but config.admin_password and possibly also config.admin_auth_token are wrong, so they won't work?

Koshmaar (approved):

utils/standalone-wizard/remote-kubernetes/remote-pub.yaml
iszulcdeepsense: Even if it's not needed, I think it's still a good practice to use non-root users and groups (just like with Dockerfiles).  So if someone have a need to change it, he probably knows what he's doing and he can modify the files before applying kubectl (Installer asks to review the resources).

utils/standalone-wizard/wizard.py
iszulcdeepsense: Yes, it is long, but there's a reason: It is downloaded via `curl` and piped to shell in a one-liner command.

utils/standalone-wizard/wizard.py
iszulcdeepsense: This is to change the default superuser password, but only for the first time. 
So I'm trying to log in with `admin:admin`. If it succeeds, I need to change the password to something more secure. Otherwise, in case of 401 Unauthorized, that means password is already changed, and don't need to do anything.
`config.admin_password` is not the actual password but the one that is supposed to be set.
I'll add a comment.

utils/standalone-wizard/wizard.py
iszulcdeepsense: Or actually there's a bash runner.sh script piped to shell (`curl runner.sh | sh`), which in turn downloads python file (and could download them more), so maybe I'll devise something better

utils/standalone-wizard/remote-kubernetes/remote-pub.yaml
Koshmaar: You're right, if somebody wants to deploy RT to his specific cluster, it's expected that he will need to change few things.

utils/standalone-wizard/wizard.py
Koshmaar: Yeah, sounds like an option.

PR #367, opened 2023-11-03
Fix reviving dead jobs

iszulcdeepsense:
Print warning in case of duplicated trash job entries due to concurrent race condition


lifecycle/lifecycle/job/models_registry.py
```diff
@@ -218,8 +233,10 @@ def create_trashed_job(job_dto: JobDto) -> models.TrashJob:
         infrastructure_target=job_dto.infrastructure_target,
         age_days=age_days,
     )
-    new_job.save()
-    return new_job
+    try:
+        new_job.save()
+    except IntegrityError:
+        logger.warning(f'Trash Job already exists with ID={job_dto.id}')
```
Koshmaar, line 239: I wonder about upping the level to error. Also, logging more info like job name, version and creation/modification time, for easier debugging. I understand you solved one issue in this PR, but maybe there will be another issue in future.

Koshmaar (approved):

PR #380, opened 2023-11-28
Deploy new jobs through Dashboard

iszulcdeepsense:
- A job can be deployed through the Dashboard.
  To submit a job manifest, go to the *"Deployments"* tab and then *"Deploy a new job"* .
  Note that secret variables are not supported in this manner; instead, use the command line client.
- A new tab called *"Deployments"* displays a list of recent deployment attempts.
  After selecting one, you may view deployment data such as status, failure cause, duration, and so on.
  In case of trouble with deployment, you can share this link with a support team.

Screenshot from 2023-11-28 11-09-56


dashboard/dashboard-front/src/components/jobs/JobDetails.vue
```diff
@@ -118,7 +105,7 @@ function openJobGrafanaDashboard(job: JobData) {
 }
 
 function editJobManifest(job: JobData) {
-    manifestDialogRef.value?.openDialog(job, manifestYaml.value)
+    manifestDialogRef.value?.openDialog(job, job.manifest_yaml || '')
```
mpolomdeepsense, line 108: I think it would be preferred to use `??` instead of `||` in this case.

dashboard/dashboard-front/src/components/jobs/ManifestView.vue
```diff
@@ -0,0 +1,21 @@
...
+import 'highlight.js/styles/github.css'
+
+hljs.registerLanguage('yaml', hljs_yaml)
+
+const props = defineProps({
+    manifestYaml: { type: String, required: false },
+})
+
+const manifestHtml: Ref<string> = computed(() => {
+    const manifestYamlN = props.manifestYaml || ''
```
mpolomdeepsense, line 14: `??` instead of `||`

dashboard/dashboard-front/src/components/jobs/deployment/DeployForm.vue
```diff
@@ -0,0 +1,96 @@
...
+import { useRouter } from 'vue-router'
+import yaml from 'js-yaml'
+import { apiClient } from '@/services/ApiClient'
+import { toastService } from '@/services/ToastService'
+import { progressService } from "@/services/ProgressService"
+
+const yamlManifestRef: Ref<string> = ref('')
+const gitUsername = ref('')
+const gitPassword = ref('')
+const forceEnabled: Ref<boolean> = ref(false)
```
mpolomdeepsense, line 12: Won't it automatically assume you want to use string in line 9 and boolean in line 12? Ref<> shouldn't be required.

dashboard/dashboard-front/src/components/jobs/deployment/DeployForm.vue
```diff
@@ -0,0 +1,96 @@
...
+        if (typeof manifestDict !== 'object') {
+            throw new TypeError('Expected mapping object')
+        }
+    } catch(err: any) {
+        toastService.showErrorDetails('Invalid Manifest YAML', err)
+        return
+    }
+
+    let gitCredentials: CredentialsModel | null = null
+    if (gitUsername.value !== '' || gitPassword.value !== '') {
```
mpolomdeepsense, line 35: `!gitUsername.value || !gitPassword.value` should also work.

Does this check make sense? I might be wrong but you use both the values if the conditional is true and yet you use `||`. Wouldn't `&&` make more sense?

dashboard/dashboard-front/src/components/jobs/deployment/DeploymentDetails.vue
```diff
@@ -0,0 +1,126 @@
...
+import { toastService } from '@/services/ToastService'
+import { apiClient } from '@/services/ApiClient'
+import {formatDuration, timestampToLocalTime} from '@/utils/time'
+import {type DeploymentDto} from '@/utils/api-schema'
+import TimeAgoLabel from "@/components/jobs/TimeAgoLabel.vue"
+import ManifestView from "@/components/jobs/ManifestView.vue"
+
+const route = useRoute()
+const deploymentId = route.params.deploymentId
+const deployment: Ref<DeploymentDto | null> = ref(null)
```
mpolomdeepsense, line 13: It doesn't make much difference but you could also use generics of `ref` function instead of importing `Ref` type. `const deployment = ref<DeploymentDto | null>(null)`

dashboard/dashboard-front/src/components/jobs/deployment/DeploymentsList.vue
```diff
@@ -0,0 +1,73 @@
+<script setup lang="ts">
+import { ref, type Ref, onMounted } from 'vue'
+import { toastService } from '@/services/ToastService'
+import { apiClient } from '@/services/ApiClient'
+import { timestampPrettyAgo, timestampToLocalTime } from '@/utils/time'
+import { type DeploymentDto } from '@/utils/api-schema'
+
+const deploymentsData: Ref<DeploymentDto[]> = ref([] as DeploymentDto[])
```
mpolomdeepsense, line 8: Same here. Would even look nicer since you wouldn't have to typecast an array.

dashboard/dashboard-front/src/components/jobs/deployment/DeploymentsList.vue
```diff
@@ -0,0 +1,73 @@
...
+        <q-card-section class="q-pb-none">
+            <div class="full-width row wrap justify-end">
+                <q-btn color="primary" push label="Deploy a new job" icon="add" :to="{name: 'deploy-job'}" />
+            </div>
+        </q-card-section>
+
+        <q-card-section>
+            <q-list bordered separator class="rounded-borders">
+
+                <q-item v-if="deploymentsData.length == 0" class="text-grey-6">(empty)</q-item>
```
mpolomdeepsense, line 45:
```suggestion
                <q-item v-if="!deploymentsData.length" class="text-grey-6">(empty)</q-item>
```

dashboard/dashboard-front/src/components/jobs/deployment/DeploymentsList.vue
```diff
@@ -0,0 +1,73 @@
...
+
+                <q-item clickable v-ripple
+                        v-for="item in deploymentsData"
+                        :to="{name: 'deployment-details', params: {deploymentId: item.id}}">
+
+                    <q-item-section avatar>
+                        <q-spinner color="primary" size="2em" v-if="item.status == 'in_progress'"/>
+                        <q-icon color="positive" name="check_circle" v-if="item.status == 'done'"/>
+                        <q-icon color="negative" name="error" v-if="item.status == 'failed'"/>
```
mpolomdeepsense, line 54:
```suggestion
                        <q-spinner color="primary" size="2em" v-if="item.status === 'in_progress'"/>
                        <q-icon color="positive" name="check_circle" v-if="item.status === 'done'"/>
                        <q-icon color="negative" name="error" v-if="item.status === 'failed'"/>
```

dashboard/dashboard-front/src/utils/time.ts
```diff
@@ -65,3 +65,24 @@ export function timestampSecondsAgo(timestampS: number | undefined): number | nu
     const diffMs = new Date().getTime() - timestampS * 1000
     return Math.floor(diffMs / 1000)
 }
+
+export function formatDuration(timestampStart: number | undefined, timestampEnd: number | undefined): string {
+    if (!timestampStart || !timestampEnd)
+        return ''
```
mpolomdeepsense, line 71: This will return an empty string when any of the input values are `0`. It's supposed to be a timestamp so this probably has no chance of occurrence. Just making sure it's supposed to work that way.

dashboard/dashboard-front/src/components/jobs/JobDetails.vue
iszulcdeepsense: Thanks, I haven't heard about this operator before

dashboard/dashboard-front/src/components/jobs/deployment/DeployForm.vue
iszulcdeepsense: Right, it makes sense to keep it consistent

mpolomdeepsense (approved):

PR #381, opened 2023-11-28
Graceful shutdown on SIGTERM

iszulcdeepsense:


racetrack_commons/racetrack_commons/api/asgi/asgi_server.py
```diff
@@ -26,20 +28,44 @@
     'GET /metrics/ 200',
 }
 
-UVICORN_DEBUG_LOGS = False
+UVICORN_ERROR_LOGS = True
```
LookACastle, line 31: This looks like a leftover from testing.

racetrack_commons/racetrack_commons/api/asgi/asgi_server.py
```diff
@@ -26,20 +28,44 @@
...
+        app=app,
+        host=http_addr,
+        port=http_port,
+        log_level="debug",
+        reload=use_reloader,
+        timeout_graceful_shutdown=3,
+    )
+    server = ManagableServer(config)
+
+    def signal_handler(sig, frame):
```
LookACastle, line 56: It might be prudent to have the signal handler be a bit more extendable if it isn't too bothersome. Maybe someone wants `SIGHUP` support or something someday. For now the signal handler seems to assume all signals want to terminate it.

There's a decent argument that we are not going to need it though, so I trust your judgement here.

racetrack_commons/racetrack_commons/api/asgi/asgi_server.py
iszulcdeepsense: No, I left that on purpose. It turns out that uvicorn sometimes spits out something useful, especially when it receives a signal and shuts down. Plus, it's not a good idea to keep the potential errors silent.

racetrack_commons/racetrack_commons/api/asgi/asgi_server.py
iszulcdeepsense: I see, maybe I'll just call it `shutdown_signal_handler` since it's only handling SIGINT and SIGTERM right now.

PR #390, opened 2023-12-06
Use podAntiAffinity to spread replicas between different nodes

iszulcdeepsense:


iszulcdeepsense:
@LookACastle 
> Code all looks fine, but I can't find it written anywhere _why_ we're doing this. There's no description in the issue (unless I missed it), and there's nothing here either. I feel like it'd be nice to have a short little blurb.

Good point, sorry about this negligence. I've added description:

> On one of our Kubernetes clusters, we spotted that 3 Lifecycle pods were all running on the same node, despite the fact that there were 3 nodes available. This is bad in terms of potential failure of this node and unnecessary downtime.
> 
> We should use [Pod's affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) to distribute pods equally between the nodes.

LookACastle (approved):

PR #395, opened 2023-12-11
Keep secret vars apart from regular env vars

iszulcdeepsense:
Secret vars are now passed separately to a job apart from regular env vars so that they can be distinguished by a plugin and handled in a different way.

Infrastructure targets can now mount secret vars to a job regardless of environment variables.
Secret variables are passed as a separate argument, they're no longer mixed with regular env vars.
Infrastructure plugins have to be updated accordingly due to interface change.
Specifically, `deploy_job` function of `JobDeployer` class now has
a new argument `runtime_secret_vars: Dict[str, str]` with secret env vars for a job.


lifecycle/lifecycle/deployer/provision.py
```diff
@@ -57,15 +57,18 @@ def provision_job(
         family = create_job_family_if_not_exist(manifest.name)
         family_dto = job_family_model_to_dto(family)
 
-        runtime_env_vars = merge_env_vars(manifest.runtime_env, secret_runtime_env)
         build_env_vars = merge_env_vars(manifest.build_env, secret_build_env)
-        runtime_env_vars = hide_env_vars(runtime_env_vars, build_env_vars)
+        all_runtime_vars = merge_env_vars(manifest.runtime_env, secret_runtime_env)
+        runtime_env_vars = manifest.runtime_env or {}
```
Koshmaar, line 62: I'm thinking if naming could be more consistent here. Because if `build_env + secret_build_env = build_env_vars`, then
`runtime_env + secret_runtime_env` should equal `runtime_env_vars`..? But such name is used below. So maybe it's better to rename `build_env_vars` to `all_build_vars`.

lifecycle/lifecycle/deployer/provision.py
```diff
@@ -57,15 +57,18 @@ def provision_job(
...
+        for build_var in build_env_vars.keys():
+            if build_var not in all_runtime_vars:
+                runtime_env_vars[build_var] = ''  # Clear out unused build env vars
```
Koshmaar, line 65: This comment would be much more useful if instead of what is happening (it's visible), it explained why it is needed.

sample/python-env-secret/README.md
```diff
@@ -1,12 +1,13 @@
...
-```
-1. Edit `.env` file and fill out your secrets.
-1. Ensure the secret file is ignored by git.
-1. Deploy the sample.
+    ```
+    cp build.env.template build.env
+    cp runtime.env.template runtime.env
+    ```
+2. Edit `build.env` and `runtime.env` files and fill out your secrets.
+3. Ensure the secret file is ignored by git.
```
Koshmaar, line 9: I believe here the terminology is a bit overgrown. In line 1 there's `env file`, which is the same as `secret file` in line 3. Also, for someone with k8s background, it's a bit misleading to read about env vars and secret vars, as they are differently handled there. But here it seems env vars (like some service url) co-exist with secret vars (like auth tokens) in the same file. Not sure what do to here, just sharing my doubts.

docs/CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Infrastructure targets can now mount secret vars to a job regardless of environment variables.
+  Secret variables are passed as a separate argument, they're no longer mixed with regular env vars.
+  Infrastructure plugins have to be updated accordingly due to interface change.
+  Specifically, `deploy_job` function of `JobDeployer` class now has
+  a new argument `runtime_secret_vars: Dict[str, str]` with secret env vars for a job.
```
Koshmaar, line 13: Can you imagine a way to make this change non-breaking? So that all plugins don't have to be updated instantly once a new version rolls in (which is operationally difficult), but lazily when there's time?

docs/development/developing-plugins.md
```diff
@@ -123,3 +123,125 @@ def post_job_delete(self, job: JobDto, username_executor: str = None):
...
+        family: JobFamilyDto,
+        containers_num: int = 1,
+        runtime_secret_vars: Dict[str, str] | None = None,
+    ) -> JobDto:
+    """Deploy a Job from a manifest file"""
+```
+
+- `delete_job` - Delete a Job based on its name
+```python
+    def delete_job(self, job_name: str, job_version: str):
```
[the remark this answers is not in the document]
iszulcdeepsense: Right, I'll add type hint

lifecycle/lifecycle/deployer/provision.py
iszulcdeepsense: Good idea, will rename to `all_build_vars`

sample/python-env-secret/README.md
iszulcdeepsense: Alright, I'll call it "secret files". You're right, they no longer have anything to do with env vars (except being kept in `.env` files format).

docs/CHANGELOG.md
Koshmaar: Yes, one way to do that would be to inspect a method signature and check if it has this argument defined. 

I know :) it's not ideal codewise, but it avoids much bigger operational problem (at least I think it will be problem, as there are dozens of plugins - how can you re-upload all right after starting newer RT version?). When all plugins are updated, such checking code can be removed - so ideally, even few days later. Last but not least, Python is good in meta inspections of code.

iszulcdeepsense: Alright, you convinced me. I'll use inspection to make plugins backward-compatible.

iszulcdeepsense: Take a look at https://github.com/TheRacetrack/racetrack/pull/395/commits/9c2e0bced05a5b20bfbff6345962ddd691b55830 . It wasn't that bad.

Koshmaar (approved):

racetrack_commons/racetrack_commons/plugin/call.py
```diff
@@ -0,0 +1,30 @@
+from inspect import getfullargspec
+from typing import Callable, Dict, Any
+
+from racetrack_client.log.logs import get_logger
+
+logger = get_logger(__name__)
+
+
+def safe_call(function: Callable, **kwargs):
```
Koshmaar, line 9: Ahh, you even made it generic. I thought about just checking for that single argument, but this is much better. Safety and future-proofness. I give you :bacon:

PR #398, opened 2023-12-20
Count jobs by job type

iszulcdeepsense:
*Administration* tab in Dashboard shows number of jobs using the particular job type or infrastructure.
  This helps to indicate which job types are actually in use and which can be deleted.
  It also shows which job types and infrastructures are provided by which plugins in a tree view.

Screenshot from 2023-12-20 10-13-02


dashboard/dashboard-front/src/components/XTooltip.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+    <q-tooltip class="bg-primary text-body2">
+        <div style="max-width: 100ch;">
+            <slot />
+        </div>
+    </q-tooltip>
+</template>
```
LookACastle, line 7: No newline at end of file (I think? Github has this odd symbol here, but inspect element calls it the no newline at end of file symbol.)

LookACastle (commented):
Looks good other than the missing newline

PR #399, opened 2024-01-04
Make Dashboard's real-time stream work with multiple replicas

iszulcdeepsense:
Currently, dashboard's auto-update feature (aka "auto-DDoS") can't connect to Lifecycle's events stream when Lifecycle runs on more than one replica. Usually it ends up with an infinite reconnection loop. Client connects to Lifecycle's socket.io server and gets the session ID, then with the next request (btw: why there are many requests being made?), the other replica handles it and doesn't know that session, thus rejecting it.
This reconnection loop was causing a lot of fuss and had to be disabled on Dashboard.

Dashboard and Lifecycle didn't use raw websockets, but socket.io library.
Now it's been switched to use use primal websocket protocol instead, which makes browser to keep just one connection to the Lifecycle server.

Dashboard's auto-update feature works with multiple Lifecycle replicas.
Live updates on the job list are enabled.


dashboard/dashboard-front/src/components/jobs/JobsListView.vue
```diff
@@ -149,56 +146,55 @@ watch(autoUpdateEnabled, () => {
...
-    })
+    ws.onerror = (err) => {
+        console.error(`events stream error: ${err}`)
+    }
+
+    ws.onclose = (event) => {
+        console.log(`disconnected from live events stream`)
+    }
+}
+
```
LookACastle, line 191: There doesn't seem to be an "on reconnect" handler anymore, is this intentional?

dashboard/dashboard-front/src/components/jobs/JobsListView.vue
```diff
@@ -149,56 +146,55 @@ watch(autoUpdateEnabled, () => {
...
+function onWebsocketEvent(event_type: string, event: any) {
+    if (event_type == 'job_models_changed') {
+        console.log('Change detected, reloading jobs')
+        const intervalMs = Math.random() * 750
+        setTimeout(fetchJobs, intervalMs) // wait random interval to avoid DDoS-ing a server
```
LookACastle, line 196: Nitpick: I had to look up what the exact return value of `Math.random()` is as someone less versed in the arts of JavaScript. Perhaps

```suggestion
        setTimeout(fetchJobs, intervalMs) // wait between 0 and 750 ms to avoid DDoS-ing a server
```

docs/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Dashboard's auto-update feature works with multiple Lifecycle replicas.
+  Live updates can now be enabled on the jobs list. It is turned off by default.
```
LookACastle, line 10: Is there a reason we change default behaviours? To me, automatic updates seem like the type of thing you'd want on by default.

lifecycle/lifecycle/event_stream/server.py
```diff
@@ -1,8 +1,15 @@
 import asyncio
 import threading
-
-import socketio
 import time
+import warnings
+
+warnings.filterwarnings(
```
LookACastle, line 6: This whole thing seems a tad cryptic, why are we ignoring a deprecation warning? Should we not update `falcon.media` if it's deprecated? If the answer is we shouldn't, then maybe adding a comment explaining why would be prudent.

dashboard/dashboard-front/src/components/jobs/JobsListView.vue
iszulcdeepsense: Yes, this is intentional. Raw websockets doesn't support reconnection handler and I didn't want to write my own in order not to end up with the infinite loop trying to reconnect and DDoS-ing a server.

lifecycle/lifecycle/event_stream/server.py
iszulcdeepsense: There's an issue which is still open https://github.com/falconry/falcon/issues/2066

docs/CHANGELOG.md
iszulcdeepsense: Well, in the past this feature generated a lot of requests, because clients were stuck in a reconnection loop and trying to get the fresh data from a server. That's why I'm a bit more careful this time. I don't want to strain the servers too much. Let's see how it goes and we can make it default again in the future.
The good news is that this option is saved in a local storage of a browser, so if you set it once, it will be turned on even after refreshing or closing the page.

PR #411, opened 2024-01-29
Implement structured logging

iszulcdeepsense:
- This is mostly a refactor to use Python logging facility  and pass additional variables as "extra".
- Additionally, few changes in Pub (which is written in Go) to produce logs in the same format.
- A new sample job in `sample/python-logging-format/entrypoint.py`.

### Added
Structured logging can be enabled by setting `LOG_STRUCTURED` environment variable to `true`.
  This will cause Racetrack services to produce logs in a JSON format.
  The default logging formatter can also be changed by jobs. 
  For more details, refer to the documentation of the specifc job type plugin.
  Take a look at the `sample/python-logging-format/entrypoint.py`
  to see how to configure your own logging formatter in jobs.

### DIFF
--- a/racetrack_commons/racetrack_commons/api/tracing.py
+++ b/racetrack_commons/racetrack_commons/api/tracing.py
@@ -5,8 +5,10 @@ from fastapi import Request
 from exceptiongroup import ExceptionGroup
 
 from racetrack_client.log.exception import exception_details
-from racetrack_client.log.logs import logger
+from racetrack_client.log.logs import get_logger
 from racetrack_client.utils.env import is_env_flag_enabled
+
+logger = get_logger(__name__)
 
 
 class RequestTracingLogger(logging.LoggerAdapter):


racetrack_commons/racetrack_commons/api/tracing.py
```diff
@@ -4,27 +4,29 @@
 from fastapi import Request
 from exceptiongroup import ExceptionGroup
 
-from racetrack_client.log.exception import get_exc_info_details
-from racetrack_client.log.logs import get_logger
-from racetrack_commons.api.debug import is_env_flag_enabled
-
-logger = get_logger(__name__)
```
ahnsn, line 11: Do you want to keep using `get_logger(__name__)` here as well?

iszulcdeepsense: Good point, I'll switch to `get_logger` factory

ahnsn (approved):

PR #417, opened 2024-02-08
Asynchronous calls to jobs

iszulcdeepsense:
### Added
- Introduced asynchronous calls to jobs. This feature lets you avoid timeouts on very long requests
  by switching to a two-step process: request a task and then check on the result.
  See [Asynchronous calls to jobs](https://github.com/TheRacetrack/racetrack/blob/412-asynchronous-calls-to-jobs/docs/user/async-job-calls.md) guide for more details.


pub/async_test.go
```diff
@@ -0,0 +1,180 @@
...
+				Body:          httpmock.NewRespBodyFromString(body),
+				Header:        http.Header{},
+				ContentLength: -1,
+				Request:       req,
+			}
+			return &resp, nil
+		})
+}
+
+func setupReplicaServer(addr string, cfg *Config, store *AsyncTaskStore) *http.Server {
```
anders314159, line 117: Very nitpicky, but I think we should call *AsyncTaskStore-parameter taskStore like in the rest of the code.

anders314159 (commented):
Looks good to me.
The only worry is using a global lock for the task store - this _might_ become a problem at some point, but until we actually see it being a problem, I don't want to implement and test something more complicated for accessing the task store.

pub/async_test.go
iszulcdeepsense: Right, renamed to `taskStore`

iszulcdeepsense:
> Looks good to me. The only worry is using a global lock for the task store - this _might_ become a problem at some point, but until we actually see it being a problem, I don't want to implement and test something more complicated for accessing the task store.

Yeah, it could become a problem when having a lot of concurrent requests. It is a "Readers–Writer lock", allowing multiple readers at no cost, but when someone writes to it, it can cause longer delays. Fortunately, writing (creating a new task) doesn't happen so often, so I think there's nothing to worry about right now.

PR #419, opened 2024-02-14
Migrate pydantic to v2

iszulcdeepsense:
### Changed
- `pydantic` package has been upgraded to version 2 to prevent from conflicts with other up-to-date packages.
  `pydantic` is one of the dependencies of the `racetrack-client` package.

### DIFF
--- a/image_builder/image_builder/api.py
+++ b/image_builder/image_builder/api.py
@@ -2,7 +2,7 @@ from typing import Any, Dict, List, Optional
 
 from fastapi import APIRouter, FastAPI
 from fastapi.responses import JSONResponse
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 from image_builder.config import Config
 from image_builder.build import build_job_image
@@ -66,7 +66,9 @@ def _setup_api_endpoints(api: APIRouter, config: Config, plugin_engine: PluginEn
         username: str = Field(examples=["admin"])
         password: str = Field(examples=["hunter2"])
 
-    class BuildPayloadModel(BaseModel, arbitrary_types_allowed=True):
+    class BuildPayloadModel(BaseModel):
+        model_config = ConfigDict(arbitrary_types_allowed=True)
+
         manifest: Dict[str, Any] = Field(
             description='Manifest - build recipe for a Job',
             examples=[{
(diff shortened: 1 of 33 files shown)


image_builder/image_builder/api.py
```diff
@@ -63,13 +63,13 @@ def _health():
 def _setup_api_endpoints(api: APIRouter, config: Config, plugin_engine: PluginEngine):
 
     class CredentialsModel(BaseModel):
-        username: str = Field(example="admin")
-        password: str = Field(example="hunter2")
+        username: str = Field(examples=["admin"])
+        password: str = Field(examples=["hunter2"])
```
anders314159, line 67: The password needs to be plaintext right? I am only seeing 'password: str = Field(examples=["*******"])'

image_builder/image_builder/config/config.py
```diff
@@ -1,9 +1,9 @@
 from typing import Optional
 
-from pydantic import BaseModel, Extra
+from pydantic import BaseModel
 
 
-class Config(BaseModel, extra=Extra.forbid):
+class Config(BaseModel, extra='forbid'):
```
anders314159, line 6: Is there a reason to keep the `extra` parameter? I am not seeing it used anywhere in the below.
https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra says that we need a model_config attribute if we plan to use 'forbid'.

anders314159: Is it somehow passed to somewhere?

lifecycle/lifecycle/config/config.py
```diff
@@ -1,10 +1,10 @@
 from typing import Optional
-from pydantic import BaseModel, Extra, validator
+from pydantic import BaseModel, field_validator
 
 from racetrack_client.utils.quantity import Quantity
 
 
-class Config(BaseModel, extra=Extra.forbid, arbitrary_types_allowed=True):
+class Config(BaseModel, extra='forbid', arbitrary_types_allowed=True):
```
anders314159, line 7: Same thought as in the other config.py comment.

image_builder/image_builder/config/config.py
iszulcdeepsense: You're right, I see I need to switch to a model_config. It's not passed anywhere, it just changes default validator, so that it doesn't accept redundant fields

PR #427, opened 2024-03-04
initial implementaion of piped input to the CLI deploy

anders314159:


racetrack_client/racetrack_client/main.py
```diff
@@ -63,9 +64,23 @@ def _deploy(
...
+        with tempfile.NamedTemporaryFile(mode='w+', dir=current_workdir, delete=False) as tmp_file:
+            tmp_file.write(piped_data)
+            file_path = tmp_file.name
+        workdir = file_path
+
     send_deploy_request(workdir, lifecycle_url=remote or _remote, force=force, build_context_method=build_context,
                         extra_vars=_parse_key_value_pairs(extra_vars))
 
+    if piped_data and file_path:
+        # Does not remove file if deploy fails.
```
iszulcdeepsense, line 82: So maybe it's better to enclose it in a try-finally block to handle errors.

racetrack_client/racetrack_client/main.py
```diff
@@ -63,9 +64,23 @@ def _deploy(
...
+    # Since most of send_deploy_request assumes a directory and a manifest file, we create one from the piped data
+    piped_data = ""
+    file_path = ""
+    if not sys.stdin.isatty() and (piped_data := sys.stdin.read()):
+        current_workdir = os.getcwd()
+        with tempfile.NamedTemporaryFile(mode='w+', dir=current_workdir, delete=False) as tmp_file:
+            tmp_file.write(piped_data)
+            file_path = tmp_file.name
+        workdir = file_path
+
```
iszulcdeepsense, line 77: Nitpick: I think it all could be moved to racetrack_client/client/deploy.py, keeping this logic away from main.py

racetrack_client/racetrack_client/main.py
```diff
@@ -63,9 +64,23 @@ def _deploy(
     extra_vars: Optional[List[str]] = typer.Option(None, '-e', '--extra-vars', help='key=value pairs overriding manifest values'),
 ):
     """Send request deploying a Job to the Racetrack cluster"""
+    # Catch if something was piped to `deploy`. Works best when the build_context is git
+    # Since most of send_deploy_request assumes a directory and a manifest file, we create one from the piped data
+    piped_data = ""
+    file_path = ""
+    if not sys.stdin.isatty() and (piped_data := sys.stdin.read()):
```
iszulcdeepsense, line 71: Does `sys.stdin.read()` blocks forever if the stdin buffer is empty?

anders314159: I think it is supposed to.

We can use a selector with a timeout of 0.5 seconds or something.

racetrack_client/racetrack_client/main.py
anders314159: I think some of it _has_ to be in main to capture piped input. I'll most of the logic to deploy.py, as I agree that main.py should be clean.

racetrack_client/racetrack_client/main.py
iszulcdeepsense: According to https://stackoverflow.com/questions/3762881/how-do-i-check-if-stdin-has-some-data, maybe that's enough to just check `not sys.stdin.isatty()`

racetrack_client/racetrack_client/main.py
```diff
@@ -1,5 +1,6 @@
 import os
 import sys
+import tempfile
```
iszulcdeepsense, line 3: Looks unused

racetrack_client/racetrack_client/client/deploy.py
```diff
@@ -65,37 +97,49 @@ def send_deploy_request(
...
-    response = parse_response_object(r, 'Lifecycle deploying error')
-    deploy_id = response["id"]
-    logger.info(f'job deployment requested: {deploy_id}')
-
+    # Creates a tmp file to store the piped input. Have to try/finally to ensure it gets deleted
     try:
-        job_url = _wait_for_deployment_result(lifecycle_url, deploy_id, user_auth, [])
+        # If piped input is available, use it as the workdir
+        tmp_file = None
+        tmp_file = catch_piped_input()
```
iszulcdeepsense, line 104:
```suggestion
        tmp_file = catch_piped_input()
```

racetrack_client/racetrack_client/client/deploy.py
```diff
@@ -65,37 +97,49 @@ def send_deploy_request(
...
+            job_url = _wait_for_deployment_result(lifecycle_url, deploy_id, user_auth, [])
+        except Exception as e:
+            raise DeploymentError(e)
+
+        logger.info(f'Job "{manifest.name}" has been deployed. Check out {job_url} to access your Job')
     except Exception as e:
-        raise DeploymentError(e)
-
-    logger.info(f'Job "{manifest.name}" has been deployed. Check out {job_url} to access your Job')
+        raise e
```
iszulcdeepsense, line 139: This `except` block is not needed then. It works the same as having only try and finally sections.

racetrack_client/racetrack_client/main.py
anders314159: I'd prefer having just the `not sys.stdin.isatty()` check, instead of the whole selector thing. The call will still block if something is piping very slowly, but that seems like a different problem.

iszulcdeepsense: Okay, that sounds good to me.

iszulcdeepsense (approved):

PR #429, opened 2024-03-04
428 cli traceback could use some newlines

anders314159:


racetrack_client/racetrack_client/log/exception.py
```diff
@@ -28,7 +28,8 @@ def exception_details(e: BaseException) -> Tuple[str, str, str]:
     """
     traceback_ex = traceback.TracebackException(type(e), e, e.__traceback__, limit=None)
     traceback_lines = list(_get_traceback_lines(traceback_ex))
-    traceback_str = ', '.join(traceback_lines)
+    traceback_lines = ['\n' + tbl for tbl in traceback_lines]
+    traceback_str = ''.join(traceback_lines)
```
iszulcdeepsense, line 32: I see your point, it's more readable. However, it was made to be one-line on purpose. When the logs are collected by external services, like Elastic or APM collector, they are being read line by line and separate lines are treated as separate logs. This might break parsing the logs correctly. Especially, if you're filtering the logs, you might lose some important information. 
So I'd leave one-line logs or force Elastic to parse multiline traceback.

PR #432, opened 2024-03-12
Retry a Job call when a job gets restarted

iszulcdeepsense:
### Changed
- Asynchronous job calls are now resilient to restarts by automatically retrying the requests,
  The restart could be either due to an upgrade in Racetrack services
  or a job crash (like an Out of Memory kill).
  Regardless of the reason, the job call will keep retrying
  until it hits the maximum number of attempts.

- Async job calls are now stored in a Postgres database (managed by means of Lifecycle service)
- Llifecycle Supervisor cleans up periodically the old records of Async Job Call models in a database
- Not every job call is being retried. Only empty responses and broken connections from a Job are qualified as retriable error, which may indicate Out of Memory kill.
- By default, max attempts is set to 2, which means Pub only retries the request once, to avoid repeating it indefinitely.


sample/python-timeout/job.py
```diff
@@ -36,6 +38,14 @@ def perform(self, mode: str = "timeout", t: int = 250) -> dict:
         elif mode == "crash":
             raise Exception("Windows12 has stopped working")
 
+        elif mode == "exit":
+            sys.exit()
+
+        elif mode == "oom":
+            for x in range(10 * 1024 * 1024):
+                self.mem.append(x)
+            return {'mem': len(self.mem)}
```
LookACastle, line 47: This is not a foolproof way to go oom. I don't know how much memory it does take up, but personally, I'd suggest `while True` rather than a `for` loop. That way it's guaranteed that it'll go oom no matter what.

There's of course a chance that what is written takes such a ridiculous amount of memory that this is merely a technicality. If so, feel free to ignore the comment.

LookACastle (commented):
It's limited how much time I could spend on the go code. It looks good at a first glance though.

docs/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Status of an async job call can be checked at endpoint: `/pub/async/task/{ID}/status`.
   See [Asynchronous calls to jobs](./user/async-job-calls.md) guide for more details.
 
+### Changed
+- Asynchronous job calls are now resilient to restarts by automatically retrying the requests,
+  The restart could be either due to an upgrade in Racetrack services
+  or a job crash (like an Out of Memory kill).
+  Regardless of the reason, the job call will keep retrying
+  until it hits the maximum number of attempts.
+  Async job calls are stored in a database for a short period.
```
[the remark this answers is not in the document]
iszulcdeepsense: That's right. Sensitive data might be there since it keeps records of request body and response body. It actually depends which model was called and whether its requests/responses may contain sensitive data.

As a mitigation, there are 2 components making sure that a record is really deleted after "a short period". Normally, Pub service is in charge of deleting an "async job call" from a database right after retrieving it.
As a fail-safe, there's also Lifecycle Supervisor service removing periodically all obsolete records, older than 4 hours.

lifecycle/lifecycle/job/async_call.py
```diff
@@ -0,0 +1,73 @@
...
+    model.request_method = dto.request_method
+    model.request_url = dto.request_url
+    model.request_headers = dto.request_headers
+    model.request_body = new_request_body
+    model.response_status_code = dto.response_status_code
+    model.response_headers = dto.response_headers
+    model.response_body = new_response_body
+    model.attempts = dto.attempts
+    model.pub_instance_addr = dto.pub_instance_addr
+    model.retriable_error = dto.retriable_error
```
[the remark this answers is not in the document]
iszulcdeepsense: It is verbose indeed. Usually, it's one-to-one mapping (it converts JSON data into database fields). However, for some fields we need to do a basic transformation, like converting string into bytes or integer into datetime, and that makes it not really a repeatable, automated process.

Plus, I've noticed that Django ORM always makes an `UPDATE` statement on database record, even if nothing changed, so that's why I made this huge `if` check.

To make it more generic, I think I could move the definition of fields and mappers into some dictionary, but then I'd need to access the fields with [Python's `setattr`](https://docs.python.org/3/library/functions.html#setattr) by string name, which makes it more error-prone for future refactors in my opinion.

sample/python-timeout/job.py
iszulcdeepsense: This name `oom` might be misleading. This thing is actually trying to handle the different use case. I don't want it to always go OOM-killed, but for the second time it's called.

I wanted to test retrying a Job call when a job gets OOM-killed, it's hard to test it if a Job is always killed even after the restart. That's why I've made this sample job to allocate more and more memory with every request (but not too much, in order not to kill it with the first request). So once it's killed, it survives the retried attempt.

LookACastle: I would rename it or chuck in a comment explaining that then.

iszulcdeepsense: Alright, I'm gonna rename it to `malloc`

LookACastle (approved):

PR #433, opened 2024-03-13
430 force flag at deployment reuses cached image

anders314159:


image_builder/image_builder/docker/builder.py
```diff
@@ -141,13 +144,23 @@ def build_container_image(
...
 ) -> str:
     """Build OCI container image from Dockerfile and push it to registry. Return build logs output"""
     # Build with host network to propagate DNS settings (network is still isolated within an image-builder pod).
     # Job workspace is the default build context
+
+    # Ensure bad flags did not sneak in
+    build_flags_set = set(build_flags)
+    approved_flags = {'--no-cache'}
+    build_flags_set = build_flags_set - approved_flags
+    assert build_flags_set == set()
```
LookACastle, line 157: Technically correct, a bit hard to read though. I'm not sure everyone parses `set()` as the empty set, and even if they do, `build_flags_set` having to be empty is a poor name.

Either rename it a bit, or loop over every item in the set and ensure it's in the approved set. Set subtraction is admittedly cooler, but it's a bit computer sciencey.

lifecycle/lifecycle/deployer/redeploy.py
```diff
@@ -22,6 +22,7 @@ def redeploy_job(
     plugin_engine: PluginEngine,
     deployer_username: str,
     auth_subject: Optional[models.AuthSubject],
+    build_flags: list[str] = [], # Setting a default here let's us ignore build_flags when redeploying
```
LookACastle, line 25: Is there a case where we'd ever want build flags when redeploying?

anders314159: Not for now, no.

image_builder/image_builder/docker/builder.py
anders314159: There's a loop now ➰➰➰

anders314159: Hmmm I guess asserts can be skipped if someone, somehow, someway, runs racetrack with the python -O flag. I guess we should raise an exception instead of asserting?

LookACastle: If someone messes with Racetrack to that point, we don't care. At that point you could be worried they removed the line of code as well. I still find the idea of raising an error more appealing though, since it'll play nicer with logs and such.

anders314159: That too.

LookACastle: Looks like you've implemented the exception thing, I'm resolving the conversation

LookACastle (approved):

iszulcdeepsense (approved):

PR #435, opened 2024-03-15
Status endpoint for async job calls

iszulcdeepsense:


docs/user/async-job-calls.md
```diff
@@ -114,6 +116,42 @@ Content-Type: application/json
 42
 ```
 
+### 3. Checking status of the task
```
LookACastle, line 119: Just for clarification, this is provided for free by Racetrack, you don't need to update the jobtype to include this?

iszulcdeepsense: Yes, it's in the Racetrack's "core".

LookACastle (approved):

Issue #437, opened 2024-03-15
Avoid redirecting version aliases to dead Jobs


iszulcdeepsense:
When a job doesn't start successfully, it's still considered a living job, and the version alias (like `latest`) will point towards it. However, this isn't the desired outcome. We want the version alias to refer to the latest stable job, that is, a job that has been initiated without any issues.

Let's introduce new `INITIALIZING` status for the jobs to distinguish these kind of jobs from the ones that are flawlessly `RUNNING`. This way, the `ERROR` status will be reserved solely for the unexpected issues that arise after a job has been successfully initialized.

PR #444, opened 2024-04-08
Clean up the messy job after deployment fails

iszulcdeepsense:


lifecycle/lifecycle/deployer/provision.py
```diff
@@ -110,10 +114,20 @@ def on_job_alive():
     update_job(job)
 
     metric_deployed_job.inc()
-    logger.info(f'job {manifest.name} v{manifest.version} has been provisioned, deployment ID: {deployment.id}')
+    logger.info(f'job {manifest.name} {manifest.version} has been provisioned, deployment ID: {deployment.id}')
```
LookACastle, line 117: Assuming this change is because `manifest.version` already has a v in it? (Or something similar that makes it obvious it's a version number?)

If not, I'd prefer the v to stay. It makes it a little clearer in my opinion.

This comment applies to the below lines logging `{job.name} {job.version}` too.

lifecycle/lifecycle/config/config.py
```diff
@@ -68,6 +68,11 @@ class Config(BaseModel):
     # Retention in seconds of Async job call models
     async_job_call_lifetime: int = 4 * 3600
 
+    # Maximum number of seconds to wait until the job is alive in the cluster (/live endpoint responds)
+    timeout_until_job_alive: int = 15 * 60
```
LookACastle, line 72: This seems like a lot. I get the rationale about some things taking a long time to spin up - is there a way to differentiate between "not alive yet, working on it" and "errored out" though? If not, this is fine, if so, consider if it makes sense to implement that distinction to avoid a 15 minute delay between someone uploading erroneous code, and finding out it's erroneous.

lifecycle/lifecycle/deployer/provision.py
iszulcdeepsense: A while ago I noticed that `{name} v{version}` doesn't work well with version aliases like `latest` which is then transformed into `adder vlatest`, that looks stupid. It's been in many places, that's why I amend this whenever I see it.
But actually, there's no risk of it in this particular line as it makes sense for Job class, not for the Manifest class which always should have exact version. 
However, Manifest class has another version field that is vulnerable - `jobtype_version`, it may contain `latest`.
That's why I'm generally striving to unify this and I think `{name} {version}` format is the best idea anyway.

lifecycle/lifecycle/config/config.py
iszulcdeepsense: I know 15 minutes is a lot, but we observed real models that took over 8 minutes just to pull the image by Kubernetes. On the other hand, "cluster errors" happen quite rarely. If a job fails, it usually does on initialization step, which waits until the job is "ready" (with 10 minutes timeout).

> consider if it makes sense to implement that distinction

Good thinking, there's already an issue for that https://github.com/TheRacetrack/racetrack/issues/437

PR #445, opened 2024-04-09
Avoid redirecting version aliases to dead Jobs

iszulcdeepsense:
A new status `STARTING` has been introduced for a Job that is still initialising.
  This way, the `ERROR` state is reserved solely for the unexpected problems
  that arise after a job has been successfully initialised.
  It is possible to distinguish whether a job has ever run successfully.
  Thanks to this, version aliases, such as `latest` or wildcards, won't redirect to starting Jobs that are not yet ready.


racetrack_commons/racetrack_commons/entities/dto.py
```diff
@@ -7,10 +7,10 @@
 
 
 class JobStatus(Enum):
-    CREATED = 'created'  # job created but didnt appear in cluster yet
+    STARTING = 'starting'  # job created but didn't appear as ready in cluster yet, still initializing
     RUNNING = 'running'
     ERROR = 'error'  # found in cluster but has failing status
-    ORPHANED = 'orphaned'  # found in cluster but LC doesn't recall to create that
+    ORPHANED = 'orphaned'  # found in cluster but Lifecycle doesn't recall to create that
```
LookACastle, line 13: Nitpick grammar
```suggestion
    ORPHANED = 'orphaned'  # found in cluster but Lifecycle doesn't recall creating that
```

lifecycle/lifecycle/job/registry.py
```diff
@@ -133,16 +133,17 @@ def sync_registry_jobs(config: Config, plugin_engine: PluginEngine):
...
+                if job_id in cluster_jobs_map:
+                    cluster_job = cluster_jobs_map[job_id]
+                    _sync_registry_job(registry_job, cluster_job)
+                    _apply_job_notice(registry_job, available_job_types)
+                else:
+                    # job not present in Cluster
+                    if registry_job.status != JobStatus.LOST.value:
+                        logger.info(f'job is lost: {registry_job}')
+                        registry_job.status = JobStatus.LOST.value
+                        models_registry.update_job(registry_job)
```
LookACastle, line 146: The wording in this section seems to suggest this only works for cluster based infrastructure types. Is that true?

lifecycle/lifecycle/job/models_registry.py
```diff
@@ -89,7 +90,8 @@ def read_latest_wildcard_job_model(job_name: str, version_wildcard: str) -> mode
     """
     version_pattern = SemanticVersionPattern.from_x_pattern(version_wildcard)
 
-    job_queryset = models.Job.objects.filter(name=job_name)
+    job_queryset = models.Job.objects.filter(name=job_name, status__in=[
```
LookACastle, line 93: Why the double underscore for `status__in`?

lifecycle/lifecycle/job/registry.py
iszulcdeepsense: I don't really get what you mean by "cluster based infrastructure type". This code should be valid for all infrastructures. Even on a local docker engine, the job can be considered Lost if it's supposed to be running but it's really gone.

lifecycle/lifecycle/job/models_registry.py
iszulcdeepsense: Good question. That's the way of applying filters in Django. Perhaps to distinguish field name from an operator. https://docs.djangoproject.com/en/5.0/topics/db/queries/#field-lookups
https://docs.djangoproject.com/en/4.2/ref/models/querysets/#in

> Basic lookups keyword arguments take the form field__lookuptype=value. (That’s a double-underscore)

lifecycle/lifecycle/job/registry.py
iszulcdeepsense: Ah, yes, I see now. The better phrase would be "infrastructure". I used to call it "cluster" back in the old days. I will rename it

LookACastle (approved):

PR #447, opened 2024-04-15
Fetch single branch by image builder

iszulcdeepsense:
- When building a Job, Racetrack only fetches a single commit from the selected branch in the
  repository containing the Job's source code. The image builder also performs "clone" and
  "checkout" in a single step. Downloading fewer data makes the build process faster.
- `.git` folder is excluded from the job image when copying the source code.


image_builder/image_builder/git.py
```diff
@@ -1,41 +1,40 @@
...
+def fetch_repository(workspace: Path, manifest: Manifest, git_credentials: Optional[Credentials]) -> tuple[Path, str]:
     """Clone git repository to local workspace directory and return fetched project path"""
     remote_url = git_remote_with_credentials(manifest.git.remote, git_credentials)
 
     with wrap_context('cloning git repo'):
         try:
-            shell(f'GIT_TERMINAL_PROMPT=0 git clone {remote_url} {workspace.resolve()}',
-                  workdir=project_root(), print_log=False)
+            if manifest.git.branch:
+                assert '"' not in manifest.git.branch, 'illegal characters in a branch name'
```
LookACastle, line 20: I intuitively think it might be better to use `git check-ref-format --branch`. Check `man git check-ref-format`.

image_builder/image_builder/git.py
```diff
@@ -1,41 +1,40 @@
...
             raise e
 
-    assert workspace.is_dir(), f'workspace directory doesn\'t exist: {workspace}'
-
-    repo = Repo(workspace.resolve())
-    with wrap_context('checking out git branch'):
-        if manifest.git.branch:
-            repo.git.checkout(manifest.git.branch)
+    git_version = read_job_git_version(workspace)
+    shutil.rmtree((workspace / '.git').resolve())  # exclude .git from the job image
```
LookACastle, line 32: This makes me nervous. Maybe add an extra check to make sure `workspace` is properly defined? I don't think it's likely to happen, nor likely to do a whole lot of damage, but better safe than sorry when it comes to recursive removal.

image_builder/image_builder/git.py
iszulcdeepsense: Good point. However, I've just checked that 
```
git check-ref-format --branch "\"&&sl&&echo"
```
doesn't raise any alarm, but I don't want it to be injected in the next command in place of `--branch "{manifest.git.branch}"`

So I'll use both: your command and my manual validation against quotes in a branch name.

iszulcdeepsense: Or actually the main reason is that I need to validate it against command injection, so I need to check it before running any command, including `git check-ref-format`

image_builder/image_builder/git.py
iszulcdeepsense: Something like this
```
    git_path = (workspace / '.git').resolve()
    assert git_path.as_posix().startswith(str(project_root() / 'workspaces'))
```
 will make you reassured?

LookACastle (approved):

PR #460, opened 2024-05-06
Make connection refused error retriable in async job calls

iszulcdeepsense:
### Changed
- "Connection refused" error is now retriable in async job calls (will be retried automatically).


anders314159 (approved):
Looks good to me

PR #461, opened 2024-05-13
Deployment getting stuck at finalizing the build phase

iszulcdeepsense:
### Fixed

Long deployments (over 20 minutes) no longer get stuck in the final stage.

This is done by switching to Server-Sent Events, sending heartbeat messages to keep alive the connection during long requests.

Plus, couple of improvements in FastAPI ASGI app to clean up the request properly (and show logs) if a client disconnects.

### DIFF
--- a/dashboard/requirements.txt
+++ b/dashboard/requirements.txt
@@ -1,2 +1,1 @@
-httpx==0.24.0
 markdown==3.4.1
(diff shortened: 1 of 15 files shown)


LookACastle (approved):
Looks good. Is the root cause confirmed to be fixed by the heartbeat thing?

Even if not, this looks useful and ready to merge.

iszulcdeepsense:
> Looks good. Is the root cause confirmed to be fixed by the heartbeat thing?
> 
> Even if not, this looks useful and ready to merge.

Yes, I've checked it on the test cluster and it worked

PR #462, opened 2024-05-16
Verify if the submitted manifest YAML is consistent with the git repository

iszulcdeepsense:
### Added
- Submitted manifest YAML is compared with the one in the job's repository.
  In case of incosistencies, the deployment is aborted.
  This only happens if the optional setting `verify_manifest_consistency` is turned on in the Image Builder.


anders314159:
Looks good to me.
Here's what I tested:
VMC = `verify_manifest_consistency: false`/`verify_manifest_consistency: true` in the image_builder config.
local/git is `--build-context=local`/`--build-context=git`.
`no complain` = job is built and pushed. `complain` = it complains about diff in job.yaml vs job.yaml on git.

job.yaml                | VMC=False, local | VMC=True, local | VMC=False, git     | VMC=True, git |
------------------------|------------------|-----------------|--------------------|---------------|
No diff                 | no complain      | no complain     | no complain        | no complain   |
Diff                    | no complain      | no complain     | no complain        | complain      |
piped input, no diff    | no complain      | no complain     | no complain        | no complain   |
piped input, diff       | no complain      | no complain     | no complain        | complain      |

Although I am wondering if `VMC=True` should check for changes even if `--build-context=local`, or maybe it should print an info line saying something like "VMC=True, but --build-context=local, so not checking changes against git" - but worded more smartly'er than that.

anders314159 (approved):

iszulcdeepsense:
@anders314159 
> Although I am wondering if `VMC=True` should check for changes even if `--build-context=local`, or maybe it should print an info line saying something like "VMC=True, but --build-context=local, so not checking changes against git" - but worded more smartly'er than that.

Right, it makes sense not to check it for local build context, even though it should always pass having the same local file in the image builder.

image_builder/image_builder/build.py
```diff
@@ -65,7 +65,7 @@ def build_job_image(
 
             workspace, repo_dir, git_version = prepare_workspace(workspaces_path, manifest, git_credentials, build_context, deployment_id)
 
-            if config.verify_manifest_consistency:
+            if config.verify_manifest_consistency and not build_context:
```
anders314159, line 68: For the racetrack CLI:
Should it be `and build_context==BuildContextMethod.git` or something like that?
I think the above `and not build_context` is always false for the racetrack [CLI](https://github.com/TheRacetrack/racetrack/blob/60e15ccb21da71a68dc14af6b8add1198bdc85c8/racetrack_client/racetrack_client/main.py#L62), since it selects "[BuildContextMethod.default](https://github.com/TheRacetrack/racetrack/blob/60e15ccb21da71a68dc14af6b8add1198bdc85c8/racetrack_client/racetrack_client/client/deploy.py#L44)" by default.
Not sure if the build_context is set, or not set, elsewhere in the project and might be None/False.

iszulcdeepsense: In the image builder `build_context` parameter is something different. It's no longer an enum describing a type of the build context, but it's the actual compressed build context tarball, represented as bytes. It's not empty when `build_context_method == BuildContextMethod.local`, thus checking `not build_context` is enough.

anders314159: cool thanks

PR #471, opened 2024-06-04
Provide its own Dockerfile by a Job

iszulcdeepsense:
### Changed
The `job_types` method's format for plugins has been changed
  to support providing Dockerfiles directly by the Jobs.
  See [Developing plugins guide](https://github.com/TheRacetrack/racetrack/blob/470-provide-dockerfile-from-a-job-repository/docs/development/developing-plugins.md#job_types).
  Previous format is still supported for backward compatibility.

### DIFF
--- a/docs/development/plugins-job-types.md
+++ b/docs/development/plugins-job-types.md
@@ -332,10 +332,10 @@ ENV DEPLOYED_BY_RACETRACK_VERSION "{{ deployed_by_racetrack_version }}"
 
 ```python
 class Plugin:
-    def job_types(self) -> dict[str, list[str]]:
+    def job_types(self) -> dict[str, dict]:
         """
         Job types provided by this plugin
-        :return dict of job type name (with version) -> list of images: dockerfile template path relative to a jobtype directory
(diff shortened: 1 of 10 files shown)


racetrack_commons/racetrack_commons/deploy/job_type.py
```diff
@@ -15,22 +15,29 @@
 logger = get_logger(__name__)
 
 
+@dataclass
+class JobTypeImageDef:
+    source: str  # either 'jobtype' or 'job' depending on location of the Dockerfile
```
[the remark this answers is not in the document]
iszulcdeepsense: Right, I'll turn into an enum.

docs/development/plugins-job-types.md
anders314159, line 1: We should probably mention that using base_image is deprecated - unless the base_image on L23 is different from the other base_images in the code.

iszulcdeepsense: Good find, I'll remove this occurrence

anders314159 (approved):

anders314159:
Looks good to me.

Issue #472, opened 2024-06-05
Lifecycle context deadline exceeded


iszulcdeepsense:
(issue originally reported by Anna)

It looks like a performance issue in Lifecycle.

## Job logs

> failed to call job "*** latest" by ***:
> Server error '500 Internal Server Error' for url 
> 'http://pub:7005/pub/job/***/latest/api/v1/perform'
> For more information check: https://httpstatuses.com/500: 
> {"error":"Proxy request error: Getting job details: GET request to 
> Lifecycle: Get 
> \"http://lifecycle:7002/lifecycle/api/v1/auth/can-call-job/***/latest/api/v1/perform\":
> context deadline exceeded (Client.Timeout exceeded while awaiting 
> headers)","requestId":"***","status":"Internal
> Server Error"}

## Lifecycle logs

Din't find any errors in logs. Tracking down the logs by tracing ID, it gives multiple calls of Lifecycle endpoint (for some reason the same tracing ID is being used multiple times).

First request has been processed in 255 ms.
However, next call (which supposed to be the same) has been stuck for **73 s** (which may have caused the timeout in the Pub client).

## Lifecycle metrics

During that time, there was a noticeable spike of active requests and active threads, while maintaining the ordinary number of requests per second and CPU usage.

PR #475, opened 2024-06-11
Lifecycle context deadline exceeded

iszulcdeepsense:
Couple of improvements to stability and observability

### Added
- Lifecycle now provides more metrics to monitor its health.
  In particular, it tracks the number of established database connections
  and the duration of fetching Job data from the database.
  To improve logging, Lifecycle now displays more details about cancelled requests.
  ([#472](https://github.com/TheRacetrack/racetrack/issues/472))

### Changed
- The Pub-Lifecycle client now has a timeout of 30 seconds.
- The HTTP thread pool of Lifecycle has been increased to 60 threads.
  This setting is configurable through the config file.

### DIFF
--- a/lifecycle/lifecycle/server/metrics.py
+++ b/lifecycle/lifecycle/server/metrics.py
@@ -35,6 +35,23 @@ metric_event_stream_client_disconnected = Gauge(
     "Total number of disconnections from the event stream",
 )
 
+metric_database_connection_opened = Counter(
+    'lifecycle_database_connection_opened',
+    'Number of times connection to a database has been opened',
+)
+metric_database_connection_failed = Counter(
+    'lifecycle_database_connection_failed',
+    'Number of times connection to a database has failed',
+)
+metric_database_connection_closed = Counter(
+    'lifecycle_database_connection_closed',
+    'Number of times connection to a database has been closed',
+)
+metric_database_cursor_created = Counter(
+    'lifecycle_database_cursor_created',
+    'Number of times database cursor has been created',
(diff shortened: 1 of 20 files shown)


lifecycle/lifecycle/server/metrics.py
```diff
@@ -67,15 +91,26 @@ def collect_active_threads_metric() -> Iterator[Metric]:
...
     metric_name = 'lifecycle_tcp_connections_count'
-    tcp_connections = collections.Counter(p.status for p in psutil.net_connections(kind='tcp'))
     prometheus_metric = GaugeMetricFamily(metric_name, 'Number of open TCP connections by status')
     for status, count in tcp_connections.items():
         prometheus_metric.add_sample(metric_name, {
             'status': status,
         }, count)
     yield prometheus_metric
 
+    remote_ports = [c.raddr.port for c in net_connections if c.status == 'ESTABLISHED' and c.raddr.port < 10000]
```
anders314159, line 104: What ports are above 10000 that we don't want metrics for?

Also the '10000' could be a variable as we use it below too.

iszulcdeepsense: That's a good question. Initially, I didn't have that limit, but then I started seeing some random ports. Perhaps, it's connecting to its own local services. I was worried about Prometheus metrics exploding in its cardinality, that's why I've set this limit.
I think it's reasonable, cause I guess there is some sort of a common sense that the important apps use 4 digit ports.

PR #478, opened 2024-06-17
Support deployment warnings

anders314159:
First pass at supporting deployment warnings.

Added a new field 'warnings' to deployment schema, racetrack CLI now prints deployment warnings, and added a verify_manifest_consistency-warning.

Notes and questions:
1. I mostly followed the implementation for deployment phases. 
2. How are DB migrations done in live environments?
3. What other places should put warnings in the DB?


image_builder/image_builder/verify.py
```diff
@@ -12,11 +12,12 @@
...
 
-def verify_manifest_consistency(submitted_yaml: str, workspace: Path, repo_dir: Path):
+def verify_manifest_consistency(submitted_yaml: str, workspace: Path, repo_dir: Path) -> str | None:
     repo_file = _find_workspace_manifest_file(workspace, repo_dir)
     if not repo_file:
-        logger.warning("Can't find manifest file in a Job repository to verify it with submitted YAML")
-        return
+        warning = "Can't find manifest file in a Job repository to verify it with submitted YAML"
+        logger.warning(warning)
+        return warning
```
iszulcdeepsense, line 20: Can't add comment to unchanged lines, so I'll add it here:
Since this function returns warnings instead of raising errors, I think this:
```python
raise RuntimeError('Submitted job manifest is not consistent with the file found in a repository. '
                           'Did you forget to do "git push"? '
                           f'Difference: {difference}')
```
 is now supposed to be
```python
return 'Submitted job manifest is not consistent with the file found in a repository. '
                           'Did you forget to do "git push"? '
                           f'Difference: {difference}'
```

lifecycle/lifecycle/endpoints/deploy.py
```diff
@@ -126,3 +126,14 @@ def _update_deployment_phase(deploy_id: str, payload: DeploymentPhase, request:
...
+
+
+    class DeploymentWarnings(BaseModel):
+        warnings: str = Field(description='deployment warnings')
+
+    @api.put('/deploy/{deploy_id}/warnings')
+    def _update_deployment_warnings(deploy_id: str, payload: DeploymentWarnings, request: Request):
+        """Update deployment's warnings"""
+        check_auth(request)
+         # TODO: _update_deployment_phase above retruns save_deployment_phase which returns None. Any reason to do the same here?
```
iszulcdeepsense, line 138: If so, it does that by mistake. No reason to do that.

lifecycle/lifecycle/job/deployment.py
```diff
@@ -119,6 +119,17 @@ def save_deployment_phase(deployment_id: str, phase: str):
     deployment.save()
 
 
+@db_access
+def save_deployment_warnings(deployment_id: str, warnings: str):
+    try:
+        deployment = models.Deployment.objects.get(id=deployment_id)
+    except models.Deployment.DoesNotExist:
+        raise EntityNotFound(f'deployment with ID {deployment_id} was not found')
+    deployment.warnings = f'{deployment.warnings},\n{warnings}' if deployment.warnings else warnings
```
iszulcdeepsense, line 128: Maybe let's keep just one character as a delimiter for simplicity and easier parsing. Plus, warning sentence might end with a dot, and then it would look stupid.
```suggestion
    deployment.warnings = f'{deployment.warnings}\n{warnings}' if deployment.warnings else warnings
```

racetrack_client/racetrack_client/client/deploy.py
```diff
@@ -116,7 +116,8 @@ def send_deploy_request(
             job_url = _wait_for_deployment_result(lifecycle_url, deploy_id, user_auth, [])
         except Exception as e:
             raise DeploymentError(e)
-
+        if warnings := _get_warning(lifecycle_url, deploy_id, user_auth):
```
iszulcdeepsense, line 119: This looks unnecessary, you get warnings field already while getting deployment data JSON in `_wait_for_deployment_result`. So another request is redundant.

racetrack_client/racetrack_client/client/deploy.py
```diff
@@ -116,7 +116,8 @@ def send_deploy_request(
...
+            logger.warning(warnings)
```
iszulcdeepsense, line 120: I think it would be better to move it inside `_wait_for_deployment_result`, so the warning can be printed out right after it appears in the database, without waiting for the deployment to finish.

image_builder/image_builder/verify.py
anders314159: Earlier the function logged a warning, so I assumed we wanted the warning. I agree that the above should be a warning instead of an error.

lifecycle/lifecycle/job/deployment.py
anders314159: Good catch.

racetrack_client/racetrack_client/client/deploy.py
anders314159: Original reasoning was to avoid extra logic, at the cost of a request. I'll move it inside `_wait_for_deployment_result`.

anders314159:
Also remember the changelog

racetrack_client/racetrack_client/client/deploy.py
```diff
@@ -8,6 +8,7 @@
 import io
 from base64 import b64encode
 from enum import Enum
+import warnings
```
iszulcdeepsense, line 11: It looks like it's here by accident
```suggestion
```

racetrack_client/racetrack_client/client/deploy.py
```diff
@@ -164,6 +164,13 @@ def _wait_for_deployment_result(lifecycle_url: str, deploy_id: str, user_auth: s
             logger.info(f'deployment in progress: {phase}...')
         else:
             logger.info(f'deployment in progress...')
+
+    warnings = response['warnings']
+    if not warningss or warningss[-1] != warnings:  # don't print the same warnings again
+        warningss.append(warnings)
+        if warnings:
+            logger.warning(warnings)
```
iszulcdeepsense, line 172: I think it should be moved above the lines
```python
status = response['status'].lower()
    if status == 'failed':
        raise RuntimeError(response['error'])
    if status == 'done':
        return response.get('job', {}).get('pub_url', '')
```
so that we don't miss a warning if it's been applied at the very end of deployment. Note that we don't care about phases, so that's the only difference between phases and warnings.

anders314159: good call

iszulcdeepsense (approved):

PR #488, opened 2024-06-28
does this structure make sense?

anders314159:
I'm aware that I need to implement the provisioning files multiple places, but this is the overall approach I think we should use. The .yamls are messy, but I'll clean those up if this structure overall makes sense.


iszulcdeepsense (commented):
Overall it makes sense.
I think that maybe we could have no contact points specified at startup (in this public repo), so that it won't raise errors when running Racetrack off the shelf, still giving possibility to fill it in easily.

anders314159:
Dashboards (Grafana updated to 11.1.0) look fine locally, but reminder to myself to check if they still look okay when deployed.

PR #494, opened 2024-07-09
Enhance Pub's metrics to detect overloaded jobs

iszulcdeepsense:


anders314159 (approved):
Looks good to me.

Tested by calling a job with

```
    def perform(self, numbers: list[float]) -> float:
        time.sleep(20)
        os.system("kill 1")
        os._exit(1)
```

and the metrics showed up as expected.

PR #498, opened 2024-07-11
Turn automatic reconciliation into a manual button

iszulcdeepsense:


anders314159 (approved):
Looks good to me.

Tested by deploying a job in kind, then deleting the deployment with `kubectl`. The job is now 'lost'. Pressing the button brings the job back.

PR #500, opened 2024-07-25
Multiple auth tokens

iszulcdeepsense:
### Added
"External Consumers" can have multiple auth tokens to facilitate the process of rotating tokens.
  Tokens can be created and revoked from the Dashboard.
  Dashboard keeps track of the date when each token was in-use last time.

Screenshot from 2024-07-25 14-29-36


anders314159:
Tested in docker, using the UI:
1. A token work for authentication
2. Multiple tokens work for authentication at the same time
3. Revoking a token immediately removes access

anders314159 (approved):
Looks good to me

PR #504, opened 2024-08-02
490 download button sometimes work sometimes it doesnt

anders314159:
Download button now works locally, in docker and kind, and on the dev-dev cluster where this is deployed.

### DIFF
--- a/dashboard/dashboard-front/src/components/admin/AdministrationView.vue
+++ b/dashboard/dashboard-front/src/components/admin/AdministrationView.vue
@@ -142,9 +142,21 @@ function deletePlugin(name: string, version: string) {
     })
 }
 
-function downloadPlugin(name: string, version: string) {
-  window.location.href = `/api/v1/plugin/${name}/${version}/download`
-}
+// If the url has a port, we are assuming the port is lifecycle, in which case we don't need to prepend /lifecycle
+// If the url does not have a port, we need to prepend /lifecycle
+const getBaseUrl = () => {
+    const { hostname, port } = window.location;
+    if (port) {
+        return ``;
+    } else {
+        return '/lifecycle';
+    }
+};
+
+const downloadPlugin = (name: string, version: string) => {
+    const baseUrl = getBaseUrl();
+    window.location.href = `${baseUrl}/api/v1/plugin/${name}/${version}/download`;
+};
 
 function onPluginUploadFailed(err: any) {
     console.error(err)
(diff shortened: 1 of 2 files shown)


dashboard/dashboard-front/src/components/admin/AdministrationView.vue
```diff
@@ -142,9 +142,21 @@ function deletePlugin(name: string, version: string) {
...
-  window.location.href = `/api/v1/plugin/${name}/${version}/download`
-}
+// If the url has a port, we are assuming the port is lifecycle, in which case we don't need to prepend /lifecycle
+// If the url does not have a port, we need to prepend /lifecycle
+const getBaseUrl = () => {
+    const { hostname, port } = window.location;
+    if (port) {
+        return ``;
+    } else {
+        return '/lifecycle';
```
iszulcdeepsense, line 152: To make it simpler and less error-prone, I think we could always use prefixed `/lifecycle` base url, which is always accessible, even on localhost.

docs/CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pub's metrics has been revised to detect overloaded jobs easier.
   ([#493](https://github.com/TheRacetrack/racetrack/issues/493))
 
+### Fixed
+- Download button works on clusters.
```
iszulcdeepsense, line 28:
```suggestion
- Download button works on Kubernetes clusters.
```

iszulcdeepsense (changes requested):

dashboard/dashboard-front/src/components/admin/AdministrationView.vue
anders314159: I tried that, it didn't work, hence the extra logic.
image

iszulcdeepsense: I owe you an explanation and I told you the wrong thing at first.

7103 is a port of Dashboard. So this works: http://127.0.0.1:7103/dashboard
Lifecycle runs on port 7102 and this address also works http://127.0.0.1:7102/lifecycle/

When you make something on Dashboard UI, it goes through Dashboard's backend (which forwards it to Lifecycle):
Dashboard UI (web app) -> Dashboard API -> Lifecycle API
This is because browsers are pretty restrictive about CORS and it's best for the web app to make a request to the same place it came from.

Thus, your base URL should be `/dashboard`

iszulcdeepsense (approved):

PR #506, opened 2024-08-16
multiple kubernetes infrastructure target plugins

anders314159:
This overwrites other infrastructure plugins with the same name.

Now we need to inform the racetrack_client and the webinterface when this happens.


anders314159:
If the infrastructure plugin already exists and the replace option is not set, an error is shown both in the CLI and using the UI.

iszulcdeepsense (approved):

PR #510, opened 2024-08-19
Support sslmode and sslrootcert in Postgres connector

iszulcdeepsense:


anders314159 (approved):
Looks good to me. The logs show that the connections are using tls.

PR #511, opened 2024-08-20
add warning if deprecated  property is used

anders314159:


racetrack_client/racetrack_client/manifest/validate.py
```diff
@@ -31,6 +31,9 @@ def load_validated_manifest(
     with wrap_context('validating manifest'):
         validate_manifest(manifest)
 
+    if manifest.lang:
+        logger.warning(f'`lang:` property is deprecated. Use `jobtybe:` instead.')
+
```
iszulcdeepsense, line 36: We've got other deprecated properties that could be taken into account as well:
https://github.com/TheRacetrack/racetrack/blob/beb7b5375bf2831483ac6c6490abff3d1cf26047/racetrack_client/racetrack_client/manifest/manifest.py#L82C5-L85C70

iszulcdeepsense (approved):

racetrack_client/racetrack_client/manifest/validate.py
anders314159: Are we keeping track of deprecations multiple places? There's also https://github.com/TheRacetrack/racetrack/blob/29d125cb9ffdd2d24b1a908a7b47773915294b18/racetrack_client/racetrack_client/manifest/schema.json

iszulcdeepsense: Looks like we do and need to keep them 2 in sync. This whole file is a duplicate of manifest.py. I'd say schema.json is a resultant file, whereas manifest.py is a source of truth.

PR #525, opened 2024-09-24
Informative error about crash during Job initialization

iszulcdeepsense:


anders314159:
Looks good to me, but how do I actually make a job crash during initialization?

If I add:
```
    def __init__(self):
        import time, sys
        time.sleep(10)
        sys.exit()
```
to a `adder.py`, it doesn't crash when deploying it.

iszulcdeepsense:
> Looks good to me, but how do I actually make a job crash during initialization?
> 
> If I add:
> 
> ```
>     def __init__(self):
>         import time, sys
>         time.sleep(10)
>         sys.exit()
> ```
> 
> to a `adder.py`, it doesn't crash when deploying it.

I know, I had the same problem. Finally, I did
```python
os._exit(1)
```
and it crashed just "fine".

PR #566, opened 2025-02-25
Add tutorial about Setting Up Racetrack with HTTPS

iszulcdeepsense:


docs/deployment/racetrack-behind-https.md
```diff
@@ -0,0 +1,141 @@
...
+Keep in mind that a properly signed certificate is usually issued for a registered DNS domain name, not an IP address.
+Working with certificates also entails setting up a correct renewal process to update them before they expire.
+
+## TLS Termination
+It's a common practice to split responsibilities into two separate tasks that can be handled by different teams:
+- Core Application Setup (using HTTP) - not accessible externally, developers don't need to worry about implementing TLS properly
+- TLS Termination at the Edge - a secure HTTPS gateway at the system's first public entry point.
+
+## Using a Self-Signed Certificate in Local Kubernetes
+To create a custom self-signed TLS certificate for use on an HTTPS server, follow these steps:
```
LookACastle, line 24: This has dependencies, listing them might be nice - even if they are quite common.

docs/deployment/racetrack-behind-https.md
```diff
@@ -0,0 +1,141 @@
...
+    ```
+
+2.  Create a Kubernetes secret:
+    ```sh
+    kubectl create secret tls tls-secret --namespace=racetrack --key tls.key --cert tls.crt
+    ```
+    
+    Note that self-signed certificates are not trusted by browsers and will trigger security warnings.
+    They are suitable for development and testing purposes, but not recommended for production environments.
+    For production use, obtain a certificate from a trusted CA. If using a self-signed certificate in a controlled environment, you can create a root CA and use it to sign server certificates, then install the root CA certificate on client machines to avoid security warnings.
```
LookACastle, line 38:
```suggestion
    Note that this will create `tls.crt` which is a self-signed certificate which is not trusted by browsers and will trigger security warnings.
    Self-signed certificates are suitable for development and testing purposes, but not recommended for production environments.
    For production use, obtain a certificate from a trusted CA. If using a self-signed certificate in a controlled environment, you can create a root CA and use it to sign server certificates, then install the root CA certificate on client machines to avoid security warnings.
```

If we change the wording like this, I feel like we make it more explicit how to use a proper cert.

docs/deployment/racetrack-behind-https.md
```diff
@@ -0,0 +1,141 @@
...
+
+Keep in mind that a properly signed certificate is usually issued for a registered DNS domain name, not an IP address.
+Working with certificates also entails setting up a correct renewal process to update them before they expire.
+
+## TLS Termination
+It's a common practice to split responsibilities into two separate tasks that can be handled by different teams:
+- Core Application Setup (using HTTP) - not accessible externally, developers don't need to worry about implementing TLS properly
+- TLS Termination at the Edge - a secure HTTPS gateway at the system's first public entry point.
+
+## Using a Self-Signed Certificate in Local Kubernetes
```
LookACastle, line 23:
```suggestion
## Example using a Self-Signed Certificate in Local Kubernetes
```

People might be liable to skip over the current title thinking it doesn't apply to them, making clear that it is an example might make it more obvious that it's relevant to them.

PR #570, opened 2025-03-06
Display Foreign Object Names in Records Table

iszulcdeepsense:
### Added
- Names of the foreign objects in the Records table are displayed next to their IDs.

Screenshot from 2025-03-06 16-52-45

### DIFF
--- a/dashboard/dashboard-front/src/components/records/RecordsList.vue
+++ b/dashboard/dashboard-front/src/components/records/RecordsList.vue
@@ -131,17 +131,18 @@ async function fetchForeignRecordNames(): Promise<void> {
     if (!primaryKeyColumn) return
     loading.value = true
     try {
+        // For each column that is a foreign key, take its values (record IDs) and enrich the foreign IDs with matching names
         const foreignKeys = tableMetadata.value.foreign_keys
         for (const column in foreignKeys) {
             if (!visibleColumns.value.includes(column)) continue
             const foreignTableName = foreignKeys[column]
-            const record_ids = pageRows.value
-                .map(row => row.fields[column])
-                .filter((it) => it != null)
+            const recordIds = pageRows.value
+                .map(row => row.fields[column]) // take the value of the particular column
+                .filter(it => it != null)
                 .map(it => String(it))
-            if (record_ids.length === 0) continue
+            if (recordIds.length === 0) continue
             const response = await apiClient.post<FetchManyNamesResponse>(`/api/v1/records/table/${foreignTableName}/names`, {
-                record_ids: record_ids,
+                record_ids: recordIds,
             } as ManyRecordsRequest)
             const idToNameMap: Map<string, string> = new Map(Object.entries(response.data.id_to_name))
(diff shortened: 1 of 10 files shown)


dashboard/dashboard-front/src/components/records/RecordsList.vue
```diff
@@ -125,13 +118,41 @@ async function fetchRecords(): Promise<void> {
...
+    loading.value = true
+    try {
+        const foreignKeys = tableMetadata.value.foreign_keys
+        for (const column in foreignKeys) {
+            if (!visibleColumns.value.includes(column)) continue
+            const foreignTableName = foreignKeys[column]
+            const record_ids = pageRows.value
+                .map(row => row.fields[column])
+                .filter((it) => it != null)
+                .map(it => String(it))
```
LookACastle, line 141: This is a bit hard to read, is the following a correct understanding?

Take the value of pageRows
For each row, take it's coloumn
For each coloumn, ensure it's not null
If it's not null, return a string representation

If I'm understanding correctly, I'm a little confused as to how you can take the coloumn of a row.

lifecycle/lifecycle/endpoints/records.py
```diff
@@ -170,6 +180,17 @@ def list_table_records(mapper: RecordMapper, payload: FetchManyRecordsRequest, t
     )
 
 
+def enrich_record_names(mapper: RecordMapper, payload: ManyRecordsRequest, table: str) -> FetchManyNamesResponse:
+    table_type = mapper.table_name_to_class(table)
+    id_to_name: dict[str, str] = {}
+    for record_id in set(payload.record_ids):
+        if record_id:
```
LookACastle, line 187: This looks odd. How can `if record_id` be meaningful in a loop over `payload.record_ids`. I don't think `payload.record_ids` should be capable of having a `record_id` that meaningfully evaluates to `False`, that seems odd. Unless it's `None`, though in that case I'd prefer `if record_id is not None` over `if record_id`

dashboard/dashboard-front/src/components/records/RecordsList.vue
iszulcdeepsense: You're nearly right. One misconception is that I'm not taking the column, but the cell (ie. `row.fields[column]` is a cell value). `pageRows` is a whole table - list of lists of cell values, so `.map(row => row.fields[column])` narrows it down to a list of values from a single column. 

> For each coloumn

I'm not taking each column, I'm focusing on one particular column.
I know it can be confusing, so I'll add a comment to it.

lifecycle/lifecycle/endpoints/records.py
iszulcdeepsense: Good question. This is because some columns can be NULLABLE in the database, and due to that, we can receive empty strings in `payload.record_ids`. `if record_id` basically checks against empty string, because for this kind of records we clearly don't want to look up the names.

LookACastle (approved):

PR #578, opened 2025-06-09
Implement password management with direct queries 

### LINKED ISSUE #569: Reimplement user management with new direct queries logic
This issue is a follow-up to [#543](https://github.com/TheRacetrack/racetrack/issues/543) and [#480](https://github.com/TheRacetrack/racetrack/issues/480) - a process of getting rid of Django.

As part of our ongoing effort to move away from Django and improve database connection management, we need to reimplement the user management functionalities in the Racetrack Dashboard using our new custom data models and direct SQL queries. This includes creating users, changing passwords, and activating accounts. It's extremely important to ensure that password encryption remains consistent with Django's current method, allowing users to continue using their existing passwords without interruption.

- Develop new data models for user management using direct SQL queries and the record mapper.
- Implement the same encryption method as Django to maintain password integrity

mateusz-twarog-deepsense:
Resolves #569


lifecycle/lifecycle/auth/validate.py
```diff
@@ -0,0 +1,26 @@
...
+    passwords_path: str = Path(__file__).resolve().parent / "common-passwords.txt"
+    
+    with open(passwords_path, "rt", encoding="utf-8") as f:
+        passwords = {x.strip() for x in f}
+
+    if password.lower().strip() in passwords:
+        raise ValidationError("This password is too common.")
+
+class ValidationError(Exception):
+    pass
```
LookACastle, line 26: No newline at end of file. Most editors shouldn't even allow this as far as I know, your settings might be off in your editor.

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

lifecycle/lifecycle/auth/validate.py
```diff
@@ -0,0 +1,26 @@
...
+    for validator in [validate_password_length, validate_numeric_password, validate_common_password]:
+        validator(password)
+
+def validate_password_length(password: str):
+    if len(password) < 8:
+        raise ValidationError("This password is too short. It must contain at least 8 characters.")
+    
+def validate_numeric_password(password: str):
+    if password.isdigit():
+        raise ValidationError("This password is entirely numeric.")
```
LookACastle, line 14: What passwords should look like exactly is more art than science, but I feel like the standard here is usually "include lower and uppercase, numbers, and special symbols" or something like that. Evaluate if you think that that would make more sense than just checking if it's a number.

I don't personally have that strong of an opinion on the matter, so if you feel like this is good enough, you can resolve this thread on your own.

lifecycle/tests/auth/test_password.py
```diff
@@ -0,0 +1,25 @@
...
+    
+    assert hash1 != hash2  # Different salts should produce different hashes
+    assert check_password(password, hash1)
+    assert check_password(password, hash2)
+
+def test_password_compatibility():
+    django_hash = "pbkdf2_sha256$600000$ToUqATye4PAvPsSe9rzUVY$eMWAQNMhb1L22JOXMI92AzSmvUtZKeTzZvbWyJMvakE="
+
+    assert check_password("admin", django_hash)
+    assert not check_password("wrong_password", django_hash)
```
LookACastle, line 25: Another missing newline, see above.

lifecycle/tests/auth/test_password.py
```diff
@@ -0,0 +1,25 @@
...
+    password = "test_password"
+    hash1 = make_password(password)
+    hash2 = make_password(password)
+    
+    assert hash1 != hash2  # Different salts should produce different hashes
+    assert check_password(password, hash1)
+    assert check_password(password, hash2)
+
+def test_password_compatibility():
+    django_hash = "pbkdf2_sha256$600000$ToUqATye4PAvPsSe9rzUVY$eMWAQNMhb1L22JOXMI92AzSmvUtZKeTzZvbWyJMvakE="
```
LookACastle, line 22: Will this always be the same? Intuitively this would be salted, no?

mateusz-twarog-deepsense:
> * This needs an issue or text in the pull request to explain what we're changing and why. Future users of `git blame` will thank us.
> * Where is `common-passwords.txt` sourced from, and do we have a plan to keep it up to date?

1. yeah description is missing, I will fix that
2. this file comes from Django which in turn sources it from another place. It seems Django kept it intact between versions for some time, I should probably at least leave note about original source somewhere.

lifecycle/lifecycle/auth/authenticate_password.py
```diff
@@ -0,0 +1,103 @@
...
+from lifecycle.database.schema.tables import User
+from lifecycle.server.cache import LifecycleCache
+from racetrack_client.log.errors import EntityNotFound
+
+from lifecycle.auth.hasher import get_hasher, get_random_string, make_password
+
+UNUSABLE_PASSWORD_SUFFIX_LENGTH = 40
+UNUSABLE_PASSWORD_PREFIX = "!"
+
+def authenticate(username=None, password=None):
```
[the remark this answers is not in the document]
mateusz-twarog-deepsense: Good point. I will need to double check all places where it is used, but those defaults are copy-pasted from Django signatures for those functions.

lifecycle/lifecycle/auth/validate.py
mateusz-twarog-deepsense: Current password verifiers are only those that were used in Racetrack so far - in theory we have also configured one that should look for similar field data in User table, but the way we called verifiers it was never actually used.
So right now this PR is only maintaining status quo. I am open to suggestions to what we want to end up with.

lifecycle/tests/auth/test_password.py
mateusz-twarog-deepsense: It is salted. Django can read salt and number of iterations from the hash, and then encode password it wants to compare to using the exact same settings. Same happens in this implementation.

lifecycle/lifecycle/auth/authenticate_password.py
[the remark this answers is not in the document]
mateusz-twarog-deepsense: defaults removed

lifecycle/lifecycle/auth/validate.py
LookACastle: If we're just maintaining status quo here, that's completely fine. We can wait with changes.

lifecycle/lifecycle/auth/authenticate_password.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        hasher = identify_hasher(encoded)
+    except ValueError:
+        # encoded is gibberish or uses a hasher that's no longer installed.
+        fake_runtime = True
+
+    if fake_runtime:
+        # Run the default password hasher once to reduce the timing difference
+        # between an existing user with an unusable password and a nonexistent
+        # user or missing hasher (similar to #20760).
```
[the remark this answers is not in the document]
mateusz-twarog-deepsense: All right, I will simplify this function a lot then.

LookACastle (approved):

PR #586, opened 2025-07-08
Create lint pipeline and fix errors

### LINKED ISSUE #581: Create lint pipeline
At this moment there is no lint enforced on pull requests. There is a command for linting ready in the makefile, though since uv is already used in the project switching to ruff might be an option.

mateusz-twarog-deepsense:
Resolves #581 and #587

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,6 +19,6 @@ jobs:
         with:
           go-version: '>=1.22'
       - name: Install dependencies
-        run: make setup-test-unit
+        run: make venv
       - name: Run lint
         run: . venv/bin/activate && make lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,6 +19,6 @@ jobs:
         with:
           go-version: '>=1.22'
       - name: Install dependencies
-        run: make venv
+        run: make setup-lint
       - name: Run lint
         run: . venv/bin/activate && make lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,7 +3,7 @@ on:
   push
 jobs:
 
-  test-unit:
+  lint:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.11'
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,24 @@
+name: lint
+on:
+  push
+jobs:
+
+  test-unit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '>=1.22'
+      - name: Install dependencies
+        run: make setup-test-unit
+      - name: Run lint
+        run: . venv/bin/activate && make lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: '>=1.22'
+          go-version: '1.22'
       - name: Install dependencies
         run: make setup-lint
       - name: Run lint
--- a/Makefile
+++ b/Makefile
@@ -78,15 +78,16 @@ install-racetrack-client:
 	( cd racetrack_client && pip install -e . )
 
 lint:
-	-python -m mypy --ignore-missing-imports --exclude 'racetrack_client/build' racetrack_client
-	-python -m mypy --ignore-missing-imports racetrack_commons
-	-python -m mypy --ignore-missing-imports --exclude 'lifecycle/lifecycle/django/registry/migrations' lifecycle
-	-python -m mypy --ignore-missing-imports image_builder
-	-python -m mypy --ignore-missing-imports dashboard
-	-python -m flake8 --ignore E501 --per-file-ignores="__init__.py:F401" \
-		lifecycle image_builder dashboard
-	-python -m pylint --disable=R,C,W \
-		lifecycle/lifecycle image_builder/image_builder dashboard/dashboard
+	python -m mypy --ignore-missing-imports --exclude 'racetrack_client/build' racetrack_client; e1=$$?;\
+	python -m mypy --ignore-missing-imports racetrack_commons; e2=$$?;\
+	python -m mypy --ignore-missing-imports --exclude 'lifecycle/lifecycle/django/registry/migrations' lifecycle; e3=$$?;\
+	python -m mypy --ignore-missing-imports image_builder; e4=$$?;\
+	python -m mypy --ignore-missing-imports dashboard; e5=$$?;\
+	python -m flake8 --ignore E501 --per-file-ignores="__init__.py:F401" \
+		lifecycle image_builder dashboard; e6=$$?;\
+	python -m pylint --disable=R,C,W \
+		lifecycle/lifecycle image_builder/image_builder dashboard/dashboard; e7=$$?;\
+	exit "$$(( e1 || e2 || e3 || e4 || e5 || e6 || e7 ))"
 
 format:
 	python -m black -S --diff --color -l 120 \
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,17 @@ setup-test-e2e:
 	( cd racetrack_commons && make setup )
 	@echo Activate your venv: . venv/bin/activate
 
+setup-lint:
+	uv venv venv &&\
+	. venv/bin/activate &&\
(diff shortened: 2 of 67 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '>=1.22'
```
LookACastle, line 20: Any reason we're pinning the python version and not the go version?

Makefile
```diff
@@ -74,19 +74,31 @@ setup-test-e2e:
...
+	python -m mypy --ignore-missing-imports --exclude 'racetrack_client/build' racetrack_client; e1=$$?;\
+	python -m mypy --ignore-missing-imports racetrack_commons; e2=$$?;\
+	python -m mypy --ignore-missing-imports --exclude 'lifecycle/lifecycle/django/registry/migrations' lifecycle; e3=$$?;\
+	python -m mypy --ignore-missing-imports image_builder; e4=$$?;\
+	python -m mypy --ignore-missing-imports dashboard; e5=$$?;\
+	python -m flake8 --ignore E501 --per-file-ignores="__init__.py:F401 lifecycle/lifecycle/event_stream/server.py:E402" \
+    lifecycle image_builder dashboard; e6=$$?;\
+	python -m pylint --disable=R,C,W \
+		lifecycle/lifecycle image_builder/image_builder dashboard/dashboard; e7=$$?;\
+	exit "$$(( e1 || e2 || e3 || e4 || e5 || e6 || e7 ))"
```
LookACastle, line 101: I'm not enough of a Makefile guru to understand what is going on here. It seems like we previously ignored non-zero exit codes, but now we exit with the cumulative `or` of every exit code? Why, and what does it even mean to `or` an exit code - is it just to check if any are truthy (I.E, nonzero)?

Also, does this even work? I was under the impression that Makefiles exited upon a nonzero return code. With the initial dash deleted, that's the case here, no?

lifecycle/lifecycle/auth/hasher.py
```diff
@@ -14,6 +21,11 @@ def make_password(password: str) -> str:
     return hasher.encode(password, salt)
 
 
+class Hash(Protocol):
+    def __call__(self, string: ReadableBuffer = b"", *, usedforsecurity: bool = True) -> hashlib._Hash:
+        ...
```
LookACastle, line 26: What does this do except be a placeholder? If it really is only a placeholder, why?

lifecycle/lifecycle/auth/users.py
```diff
@@ -66,7 +66,7 @@ def register_user_account(username: str, password: str) -> tables.User:
     grant_permission(auth_subject, AuthScope.DEPLOY_JOB.value)
 
     logger.info(f'User account created: {username}')
-    return user
+    return user_record
```
LookACastle, line 69: This looks like a functionality change rather than a linting one. Is that intentional? If so, what is it fixing? I don't mind sneaking in a bug fix in an unrelated MR, but I wanna know what is being fixed.

lifecycle/lifecycle/database/postgres/engine.py
```diff
@@ -221,9 +221,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     @classmethod
-    def connect(cls, *args, **kwargs) -> "Connection":
+    def connect(cls, *args, **kwargs) -> "PgConnection":
         try:
-            return Connection.connect(*args, **kwargs)
+            return super(PgConnection, cls).connect(*args, **kwargs)
```
LookACastle, line 226: This again seems like a non-linting change, what are we fixing?

lifecycle/lifecycle/django/registry/migrations/0023_default_superadmin.py
```diff
@@ -37,7 +37,7 @@ def create_default_superuser(apps, schema_editor):
     with DisableSignals():
         try:
             user = User.objects.get(username='admin')
-        except User.DoesNotExist:
+        except User.DoesNotExist:  # pylint: disable=E1101
```
LookACastle, line 40: When we ignore the linter, I think a comment explaining why is a good idea.

racetrack_commons/racetrack_commons/api/asgi/asgi_server.py
```diff
@@ -156,9 +156,9 @@ def format(self, record: logging.LogRecord):
 
 class NeedlessRequestsFilter(logging.Filter):
     def filter(self, record: logging.LogRecord):
-        method: str = record.args[1]
-        uri: str = record.args[2]
-        response_code: int = record.args[4]
+        method: str = record.args[1]  # type: ignore
+        uri: str = record.args[2]  # type: ignore
+        response_code: int = record.args[4]  # type: ignore
```
LookACastle, line 161: Again, if we're ignoring the linter, I think we should explain why

racetrack_commons/racetrack_commons/api/asgi/fastapi.py
```diff
@@ -101,6 +101,6 @@ def custom_openapi():
         fastapi_app.openapi_schema = openapi_schema
         return fastapi_app.openapi_schema
 
-    fastapi_app.openapi = custom_openapi
+    fastapi_app.openapi = custom_openapi  # type: ignore
```
LookACastle, line 104: Again, if we're ignoring the linter, I think we should explain why

racetrack_commons/racetrack_commons/api/metrics.py
```diff
@@ -29,7 +29,7 @@
 def setup_metrics_endpoint(api: FastAPI):
 
     metrics_app = make_wsgi_app(REGISTRY)
-    api.mount('/metrics', WSGIMiddleware(metrics_app))
+    api.mount('/metrics', WSGIMiddleware(metrics_app))  # type: ignore
```
LookACastle, line 32: Again, if we're ignoring the linter, I think we should explain why

requirements-dev.txt
```diff
@@ -1,7 +1,10 @@
...
-pylint==2.15.8
+mypy==1.16.1
+flake8==7.3.0
+pylint==3.3.7
 black==24.3.0
 twine==6.1.0
 wheel==0.38.4
+types-PyYAML==6.0.12
+types-python-dateutil==2.9.0.20250708
+types-Markdown==3.3.18
```
LookACastle, line 10: No newline at end of file

LookACastle (changes requested):

.github/workflows/lint.yml
mateusz-twarog-deepsense: Good point, I must have copy-pasted this config without thinking. In theory it should be fine since go cares about backward compatibility, but pinning makes more sense.

Makefile
mateusz-twarog-deepsense: This basically takes all exit codes and "or" on them will return 0 if all of them were 0, some other code if at least one was not 0. It works, but it's ugly - this is purely bash syntax used, since it's makefile maybe I could split it into multiple commands and last one dependent on them? I would need to experiment a bit.

lifecycle/lifecycle/auth/hasher.py
mateusz-twarog-deepsense: Ah, this one is weird. There is no type I could import from hash library that would allow me to correctly type `digest` few lines below, so I had to write correct interface myself.

lifecycle/lifecycle/auth/users.py
mateusz-twarog-deepsense: Yes, it was a bug, but since return value was always ignored anyway there is no functional change.

lifecycle/lifecycle/database/postgres/engine.py
mateusz-twarog-deepsense: Typing was incorrect here, because it involved a bit of magic, like `connect` returning  object of its own type. Functionally we are doing exactly the same thing as before, but with a syntax that makes more sense for mypy.

Makefile
LookACastle: If it works then I'm not too fussed either way. Makefiles are not exactly known for their graceful syntax

.github/workflows/lint.yml
mateusz-twarog-deepsense: pinned

lifecycle/lifecycle/django/registry/migrations/0023_default_superadmin.py
mateusz-twarog-deepsense: managed to overcome this by adding pylint_django plugin

racetrack_commons/racetrack_commons/api/asgi/fastapi.py
mateusz-twarog-deepsense: commented and narrowed the ignore

lifecycle/lifecycle/auth/hasher.py
mateusz-twarog-deepsense: commented

racetrack_commons/racetrack_commons/api/asgi/asgi_server.py
mateusz-twarog-deepsense: ended up relaxing typing since it does not matter for printing/logging

racetrack_commons/racetrack_commons/api/metrics.py
mateusz-twarog-deepsense: commented

mateusz-twarog-deepsense:
@LookACastle I've made some additional changes to lifecycle/lifecycle/database/record_mapper.py and lifecycle/lifecycle/database/type_parser.py to reduce casting and clean it up a little

LookACastle (approved):

PR #599, opened 2026-04-22
Rewrite path when proxying "latest" job requests

### LINKED ISSUE #598: Job version "latest" redirection
The way "latest" version is handled is inconsistent. 
If the url **does not** end with "latest", for example `/pub/job/sample-dockerfiled/latest/docs`, after the request is handled by pub the browser will stay at this address.
However if url ends with "latest", like `/pub/job/sample-dockerfiled/latest` behavior changes and browser gets redirected, ending with url like `/pub/job/sample-dockerfiled/0.0.1`.

Functionally that usually makes little difference, but we found use cases where it is problematic. When proxying requests to "latest" redirect causes issues as proxy shouldn't need to be able to direct requests to numbered versions.

mateusz-twarog-deepsense:
Jobs can, and often will issue redirect responses. When uses sends request to the "latest" job version, but then url is redirected by job, user would end up on a page with actual version number. This is inconsistent with usual behavior where user is always kept with urls containing "latest" as the version.

This change rewrites any redirect requests issued by jobs to keep "latest" version in them.

Resolves #598

### DIFF
--- a/pub/proxy.go
+++ b/pub/proxy.go
@@ -18,6 +18,8 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+const versionLatest string = "latest"
+
 var defaultJobProxyTransport http.RoundTripper = defaultHttpTransport()
 
 func proxyEndpoint(c *gin.Context, services *Services, jobPath string) {
@@ -94,13 +96,13 @@ func handleProxyRequest(
 	metricJobProxyRequests.WithLabelValues(job.Name, job.Version).Inc()
 
 	if !cfg.RemoteGatewayMode && jobCall.RemoteGatewayUrl != nil {
-		return handleMasterProxyRequest(c, cfg, logger, requestId, jobPath, jobCall, job, callerName, startTime)
+		return handleMasterProxyRequest(c, cfg, logger, requestId, jobPath, jobCall, job, callerName, startTime, jobVersion == versionLatest)
 	}
 
 	urlPath := JoinURL("/pub/job/", job.Name, job.Version, jobPath)
 	targetUrl := TargetURL(cfg, job, urlPath)
 
-	ServeReverseProxy(targetUrl, c, job, cfg, logger, requestId, callerName, startTime)
+	ServeReverseProxy(targetUrl, c, job, cfg, logger, requestId, callerName, startTime, jobVersion == versionLatest)
 	return http.StatusOK, nil
 }
 
@@ -126,6 +128,7 @@ func ServeReverseProxy(
 	requestId string,
 	callerName string,
 	startTime time.Time,
+	isCallingLatest bool,
 ) {
 
 	director := func(req *http.Request) {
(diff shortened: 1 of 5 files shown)


pub/proxy.go
```diff
@@ -150,6 +153,13 @@ func ServeReverseProxy(
 			if err == nil {
 				redirectUrl.Host = ""
 				redirectUrl.Scheme = ""
+
+				if isCallingLatest {
+					redirectUrl.Path = strings.Replace(redirectUrl.Path, "/"+job.Version, "/"+versionLatest, 1)
+					// clear RawPath to make sure there is no mismatch after changing Path
+					redirectUrl.RawPath = ""
```
LookACastle, line 160: What does this mean? Is it clearing `redirectUrl.RawPath` so that failures are obvious or..?

pub/proxy.go
```diff
@@ -150,6 +153,13 @@ func ServeReverseProxy(
 			if err == nil {
 				redirectUrl.Host = ""
 				redirectUrl.Scheme = ""
+
+				if isCallingLatest {
+					redirectUrl.Path = strings.Replace(redirectUrl.Path, "/"+job.Version, "/"+versionLatest, 1)
```
LookACastle, line 158: So always, even when specifically accessing a particular version, if that version happens to be the latest one, you get redirected to `/latest`. That seems counterintuitive, but I guess it's because we want the same consistent URL, and it happens to be the case that it's easier for someone to allow traffic to `/latest` than updating the version all the time if they're strict about specific endpoints in whatever solutions they have.

Is this correct?

pub/remote_gateway.go
```diff
@@ -78,6 +78,7 @@ func handleMasterProxyRequest(
 	job *JobDetails,
 	callerName string,
 	startTime time.Time,
+	isCallingLatest bool,
```
LookACastle, line 81: No strong opinion from my side, but intuitively I'd figure that the function would figure this out rather than the caller supplying the info. If it requires some weird variable shenanigans or string manipulation or something I'd probably do as you've done in the name of code aesthetics, but it's worth noting in case you agree and it's trivial.

LookACastle (commented):
No major roadblocks, just a couple of considerations that need answered for posterity's and documentation's sake.

pub/proxy.go
mateusz-twarog-deepsense: No, this substitution happens only when job is called as "latest" as in `/pub/job/adder/latest`, it won't happen if  job is called as `/pub/job/adder/1.0.0` even if 1.0.0 is currently the latest version.

pub/proxy.go
mateusz-twarog-deepsense: RawPath is a weird optional field. It is supposed to hold escaped path for some utility usage, but having it empty is valid. After changing Path it is a good idea to make sure it still matches RawPath, and cleaning up RawPath is the easiest and safest way to do it.

pub/proxy.go
LookACastle: That's even better then!

pub/remote_gateway.go
mateusz-twarog-deepsense: I can move this logic around if needed. The check we need to make is to check what version user asked for - and if it is "latest" make a substitution. If we want to move the logic, instead of passing bool here we would have a `userRequestedVersion string` or something similar.

LookACastle: Then that's not any neater. I figured it was already in the `jobPath` or something.

LookACastle (approved):

PR #603, opened 2026-05-08
Make async job results retention period configurable

### LINKED ISSUE #602: Make async job results retention period configurable

mateusz-twarog-deepsense:
Golang version bumped to 1.25 to faciliate synctest in unit tests

resolves #602

### DIFF
--- a/pub/async_store.go
+++ b/pub/async_store.go
@@ -23,6 +23,7 @@ type AsyncTaskStore struct {
 	replicaDiscovery      *replicaDiscovery
 	cleanUpTimeout        time.Duration
 	taskStorage           TaskStorage
+	cleanUpQuitChannel    chan bool
 }
 
 type AsyncTask struct {
@@ -65,7 +66,7 @@ const (
 
 var ErrAsyncTaskNotFound = errors.New("Async task not found")
 
-func NewAsyncTaskStore(replicaDiscovery *replicaDiscovery, taskStorage TaskStorage) *AsyncTaskStore {
+func NewAsyncTaskStore(replicaDiscovery *replicaDiscovery, taskStorage TaskStorage, retentionPeriod time.Duration) *AsyncTaskStore {
 	store := &AsyncTaskStore{
 		localTasks: make(map[string]*AsyncTask),
 		jobHttpClient: &http.Client{
@@ -80,10 +81,11 @@ func NewAsyncTaskStore(replicaDiscovery *replicaDiscovery, taskStorage TaskStora
 			Timeout:   5 * time.Second,
 			Transport: defaultAsyncReplicaTransport,
 		},
-		longPollTimeout:  30 * time.Second,
-		cleanUpTimeout:   125 * time.Minute,
-		replicaDiscovery: replicaDiscovery,
-		taskStorage:      taskStorage,
+		longPollTimeout:    30 * time.Second,
+		cleanUpTimeout:     retentionPeriod,
+		replicaDiscovery:   replicaDiscovery,
+		taskStorage:        taskStorage,
+		cleanUpQuitChannel: make(chan bool, 1),
 	}
 	go store.cleanUpRoutine()
 	return store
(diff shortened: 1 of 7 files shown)


pub/async_store.go
```diff
@@ -80,10 +81,11 @@ func NewAsyncTaskStore(replicaDiscovery *replicaDiscovery, taskStorage TaskStora
...
 		},
-		longPollTimeout:  30 * time.Second,
-		cleanUpTimeout:   125 * time.Minute,
-		replicaDiscovery: replicaDiscovery,
-		taskStorage:      taskStorage,
+		longPollTimeout:    30 * time.Second,
+		cleanUpTimeout:     retentionPeriod,
+		replicaDiscovery:   replicaDiscovery,
+		taskStorage:        taskStorage,
+		cleanUpQuitChannel: make(chan bool, 1),
```
mateusz-twarog-deepsense, line 88: It's a nice practice to have a way to stop goroutine running in background. In this case it's also necessary for unit test to pass since isolated environment of the synctest will not allow for any goroutines still going when test finishes.

pub/async_store.go
```diff
@@ -142,29 +144,35 @@ func (s *AsyncTaskStore) DeleteLocalTask(taskId string) {
...
-					if err != nil {
-						log.Error("Failed to delete obsolete async task", log.Ctx{
-							"taskId": task.Id,
-							"error":  err.Error(),
-						})
-					}
-				}(task)
+		select {
+		case <-s.cleanUpQuitChannel:
+			return
```
LookACastle, line 152: My understanding is that select blocks until something arrives on a channel, choosing between available channels randomly.

If that's correct this does nothing, right? If `cleanupQuitChannel` does anything, it just returns and does nothing? Why?

I guess it exits the function? But that seems like it'd be an issue too, unless there's some functionality to immediately make a new async store I'm not seeing. Because if it is just quitting the cleanup routine, then we'll just have a store with infinitely old data allowed in it. Also, shouldn't it delete the current data before quitting?

pub/config.go
```diff
@@ -1,30 +1,33 @@
...
+	RemoteGatewayToken       string        `env:"REMOTE_GATEWAY_TOKEN" envDefault:""`
+	ServiceName              string        `env:"SERVICE_NAME" envDefault:"pub"`
+	StructuredLogging        bool          `env:"LOG_STRUCTURED" envDefault:"false"`
+	ReplicaDiscoveryHostname string        `env:"REPLICA_DISCOVERY_HOSTNAME" envDefault:""`
+	AsyncMaxAttempts         int           `env:"ASYNC_MAX_ATTEMPTS" envDefault:"2"`             // Maximum number of attempts to make a job call (1 is no retry)
+	AsyncTaskRetryInterval   int           `env:"ASYNC_TASK_RETRY_INTERVAL" envDefault:"10"`     // Interval in seconds to wait before retrying a crashed job call
+	AsyncTaskRetentionPeriod time.Duration `env:"ASYNC_TASK_RETENTION_PERIOD" envDefault:"125m"` // Time for which results from async job calls will be kept
+	LifecycleCacheTTLMin     int           `env:"LIFECYCLE_CACHE_TTL_MIN" envDefault:"60"`       // Time-to-live in seconds to keep cached Lifecycle responses
+	LifecycleCacheTTLMax     int           `env:"LIFECYCLE_CACHE_TTL_MAX" envDefault:"600"`      // Time-to-live in seconds to keep cached Lifecycle responses in case of Lifecycle/Database outage
```
LookACastle, line 30: The diff looks crazy on this one. I can only spot you adding the `AsyncTaskRetentionPeriod` line - is that all that changed or did I miss something?

LookACastle (commented):
Probably approved, I'll just have you educate me a bit before I press the green button :P

pub/config.go
mateusz-twarog-deepsense: Yes, that's the only change, but gofmt insists that all types must be aligned as well as all comments in adjacent  lines.

pub/async_store.go
mateusz-twarog-deepsense: Only thing it does is that it makes it possible to exit this function when it's running in background at(almost) any time. Hanging on both this channel and timer means that it gets woken up either on timer to do its job, or on channel to exit immediately. 
In our case early exit is ever called only when the whole application is coming down, and actual job clean up is done by CancelOngoingRequests called right before. The gain for the application from this change is minimal making exit procedure a tiny bit more gracious. For our tests however we need to have a way to stop a goroutine graciously since synctest is written to ensure all gorutines exit before test ends and will panic if that is not the case.

LookACastle (approved):

PR #615, opened 2026-05-26
Support for arm64 builds

### LINKED ISSUE #614: Support for arm64 builds
Services that interact with docker, namely Image Builder and Lifecycle are hardcoded to use amd64. As a result, they cannot be built as arm64 images and at best emulated on such machines.
That makes local development on modern mac machines harder as docker-based deployment must be emulated, while local Kubernetes deployment using Kind is not possible since Kind explicitly blocks emulation. Furthermore deployment of amd64 images on arm64 nodes is a suboptimal use of resources.

A/C:

- Racetrack images can be built as arm64
- build can be configured for either architecture independently from actual machine architecture
- build defaults to the machine architecture for easy local development
- image publish still defaults to amd64 unless explicitly overridden
- docs are updated to reflect those changes - note that there is a snippet about arm64 deployment in docs/deployment/local-kubernetes-setup.md

mateusz-twarog-deepsense:
- updated lifecycle and image builder to use TARGETARCH to use correct binaries
- lifecycle and image builder now have build stage with additional dependencies
- moved "private" lifecycle image definition from separate docker file to a stage in main docker file - that allows us to not depend on public lifecycle image to be used as base
- switched to building images via buildx to allow more configuration, most importantly possibility of multi-arch builds
-  definitions of builds for pushing are now in "bake" hcl file

resolves #614

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,11 @@
 
 # docker tag of images
 TAG ?= 2.36.0
-DOCKER_REGISTRY ?= ghcr.io
-DOCKER_REGISTRY_NAMESPACE ?= theracetrack/racetrack
-GHCR_PREFIX = ghcr.io/theracetrack/racetrack
+GHCR_REGISTRY ?= ghcr.io
+GHCR_NAMESPACE ?= theracetrack/racetrack
+GHCR_PREFIX = $(GHCR_REGISTRY)/$(GHCR_NAMESPACE)
+DOCKER_REGISTRY ?= $(GHCR_REGISTRY)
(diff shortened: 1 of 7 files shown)


LookACastle (changes requested):
Code all looks fine, but some documentation somewhere should mention that this exists and describe how to use it.

mateusz-twarog-deepsense:
> Code all looks fine, but some documentation somewhere should mention that this exists and describe how to use it.

Usage is exactly the same as before, though I will as some mention that you could configure other platform targets.

LookACastle (approved):
lgtm

Issue #616, opened 2026-07-13
Change the sequence of steps for the new kubernetes-git-kustomize infrastructure target


hannaweber-deepsense:
The first job in a family can't be created with the kubernetes-git-kustomize plugin:

```
-> % racetrack deploy --build-context=local sample/python-class3
deployment error: saving job secrets: Job Family with name multiplier was not found
```

hannaweber-deepsense:
The fix is to create a family at the time an auth token with the family id is being created.

Tests:
1. Deploy the new Racetrack version (a hot fix): 2.36.0-616-1
2. Deploy update plugin, infrastructure_target=kubernetes-git-kustomize, version 1.0.1
3. Deploy jobs with infrastructure_target=kubernetes-git-kustomize
3.1. Deploy a first job in a family, i.e. the first job with the given name
3.2. Deploy a next job in a family
4. Deploy jobs with infrastructure_target=kubernetes (regression test)
4.1. Deploy a first job in a family, i.e. the first job with the given name
4.2. Deploy a next job in a family
NOTE: 
After every job deployment: get the list of jobs and verify the job is deployed and the infrastructure_target is as expected.
Tests the jobs by calling their /perform API.

PR #617, opened 2026-07-17
Create a job family before creating an auth token.

hannaweber-deepsense:
See https://github.com/TheRacetrack/racetrack/issues/616

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: setup test clean registry venv
 
 # docker tag of images
-TAG ?= 2.36.0-616-1
+TAG ?= 2.36.0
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: setup test clean registry venv
 
 # docker tag of images
-TAG ?= 2.36.0
+TAG ?= 2.36.0-616-1
```
mateusz-twarog-deepsense: If we want to release those changes we should increase patch version and make release notes for it.

hannaweber-deepsense: The release version will be set to 2.36.1 in a release branch and a next pull request.

mateusz-twarog-deepsense (approved):

